PR #17, opened 2018-03-10
feat: File globs using mattn/go-zglob library

### LINKED ISSUE #14: glob patterns for files/config_files
[This](https://github.com/goreleaser/nfpm/blob/0c6b7821812f13134ac062a82b45c863ada261a0/deb/deb.go#L124) had me thinking. Would you consider adding glob patterns to match files/folders? The obvious choice is to use the [path/filepath](https://golang.org/pkg/path/filepath/#Glob) package. However it [doesn't support the ** glob pattern](https://github.com/golang/go/issues/11862) (at least for now). Maybe one could resolve to [this library called zglob](https://github.com/mattn/go-zglob)? Any opinions?

tympanix:
Added file globs to deb and rpm packager. Since the destination for
a glob can not be a single file, the files collected by the glob are
stored in: join(dst, trim(lcp(src), src)) (read: lcp = longest common
prefix). Thereby the longest common prefix is removed from every path in
the globbed files and joined with the destination path. Implementation
is backwards compatible (i.e. non-globs will work as before)

Closes #14


pkgutil/pkgutil.go
```diff
@@ -0,0 +1,66 @@
+package pkgutil
```
caarlos0: this is not a good package name... maybe `glob`?

pkgutil/pkgutil.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	zglob "github.com/mattn/go-zglob"
+)
+
+// LongestCommonPrefix returns the longest prefix of all strings the argument
+// slice. If the slice is empty the empty string is returned
+func LongestCommonPrefix(strs []string) string {
```
caarlos0: this function could be private...

pkgutil/pkgutil.go
caarlos0: also missing docs

pkgutil/pkgutil.go
```diff
@@ -0,0 +1,66 @@
...
+func Glob(glob, dst string) (map[string]string, error) {
+	matches, err := zglob.Glob(glob)
+	if err != nil {
+		return nil, err
+	}
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("no files matched by: %s", glob)
+	}
+	files := make(map[string]string)
+	prefix := LongestCommonPrefix(matches)
```
caarlos0: i don't get why we need this..

caarlos0: oh, nevermind, just realized it...

tympanix: to be clear: by removing the longest prefix we enable the user to have the greatest control over the resulting destination path.

tympanix:
Bump. Are we good or waiting for other parts of nfpm to mature first?

caarlos0:
Hey @tympanix, I still need to test this within goreleaser and decide if it's better to put this here or there...

sorry for the delay, crazy shit happening here

tympanix:
No worries. I would personally love to have the feature here because it promotes npfm as a standalone tool as well. Win win. Looking forward to it.

caarlos0:

Issue #19, opened 2018-03-18
Add support for package scripts


chris-rock:
It would be nice to define scripts similar to fpm. The `.goreleaser.yml` could look like:

```yaml
nfpm:
  scripts:
    "verify":         "scripts/verify.sh"
    "before_upgrade": "scripts/before_upgrade_script.sh"
    "after_upgrade":  "scripts/after_upgrade_script.sh"
    "before_install"  "scripts/before_install_script.sh"
    "after_install":  "scripts/after_install_script.sh"
    "before_remove":  "scripts/before_remove_script.sh"
    "after_remove":   "scripts/after_remove_script.sh"
    "rpm_pretrans":   "scripts/rpm_pretrans_script.sh"
    "rpm_posttrans":  "scripts/rpm_posttrans_script.sh"
```

Happy to help with a PR once we agreed on a way forward

caarlos0:
LGTM

caarlos0:
> Happy to help with a PR once we agreed on a way forward

Let me know if you need help :D

chris-rock:
@caarlos0 Unfortunately I am not able to start this work for at least the next week. Please go ahead, I look for any updated before I start working

tympanix:
Any progress? I may have some time on my hands to work on this. A few remarks; if we are to use the structure above, this would imply the same script is used across multiple packagers (rpm, deb ect.). This may not be an immediate problem, but power users may want to take advantage of some explicit features in a maintainer script (e.g. the arguments passed to [debian's maintainer scripts](https://www.debian.org/doc/debian-policy/#summary-of-ways-maintainer-scripts-are-called)). Therefore we may also, in addition to the above structure, allow custom maintainer scripts on a per packager basis.

In addition, for simple use cases, I would like to propose a module system. One may be able to draw inspiration from configuration managements tools (like modules in [Ansible](http://docs.ansible.com/ansible/latest/modules/list_of_system_modules.html)). This will enable the user to specify maintainer scripts with the use of higher level directives (e.g. creating users, declaring systemd units ect.). The maintainer scripts themselves will be generated by nfpm using packager specific best practices. This relates to solving #15 as well. However I expect this to be of the least priority among the above (and certainly something for the long run).

TLDR;
 * Packager independent maintainer scripts (like proposed by OP)
 * Maintainer scripts on a per packager basis (for power users)
 * A module abstraction level for maintainer scripts (in the long run)

caarlos0:
> Any progress?

[...]

PR #35, opened 2018-06-04
docs: Add overrides section to example config

ecmrauh:
I noticed that the new feature of pre- und postinstall scripts isn't
reflected in the output of the example config file. This commit adds the
missing parts to make it clearer that this feature is available.

See #19


caarlos0:
thanks @ecmrauh !

caarlos0:

PR #38, opened 2018-09-14
Added Environment Support for version

mschneider82:
With this patch it is possible to use enviornment variables for the versions value in the yaml file:

```version: $NFPM_PACKAGE_VERSION```

The reason for this is automated builds. With Environment support for Version value I dont need to render the yaml file each time.

### DIFF
--- a/nfpm.go
+++ b/nfpm.go
@@ -42,6 +42,9 @@ func Parse(in io.Reader) (config Config, err error) {
 	if err = dec.Decode(&config); err != nil {
 		return
 	}
+	if strings.HasPrefix(config.Info.Version, "$") {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by caarlos0** [comment]
lets use `os.ExpandEnv` instead. You can even remove the `if` statement and could the use multiple variables.

Also, please add tests.

**Comment by mschneider82** [comment]
What do you think about changing the ``init`` template from:
```
version: "1.0.0" 
```

to 

```
version: "1.0.${PATCHLEVEL}"
```

to show this feature in the example config?

caarlos0:
@mschneider82 good idea, yes, or just `v$MY_APP_VERSION` or something like that

caarlos0:

PR #59, opened 2019-07-05
Adding Prefix and Group tags

AdrianWennberg:
Prefix and Group tags added for RPMs only since those tags do not exist for deb packages. 

Also defined some macros for rpmbuild to make nfpm less dependent on the build environment.

### DIFF
--- a/rpm/rpm_test.go
+++ b/rpm/rpm_test.go
@@ -100,6 +100,8 @@ func TestWithRPMTags(t *testing.T) {
 		Group:  "default",
 		Prefix: "/usr",
 	}
+	var err = Default.Package(info, ioutil.Discard)
+	assert.NoError(t, err)
 }
 
 func TestRPMVersionWithDash(t *testing.T) {
--- a/rpm/rpm_test.go
+++ b/rpm/rpm_test.go
@@ -104,6 +104,26 @@ func TestWithRPMTags(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestRPMTagsSpec(t *testing.T) {
+	var info = exampleInfo()
+	info.RPM = nfpm.RPM{
+		Group:  "default",
+		Prefix: "/usr",
+	}
+
+	vs := rpmbuildVersion{Major: 4, Minor: 15, Patch: 2}
+	golden := "testdata/spec_4.15.x.golden"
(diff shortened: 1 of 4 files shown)


caarlos0:
any thing like this for deb?

if its only for RPM, maybe prefix the config with `RPM` or inside a RPM block to make it clearer 🤔

AdrianWennberg:
The `Section` tag which is already implemented for deb-only seems to be similar to the `Group` tag for RPMs, but it does not make sense to use one tag for both since they have different naming conventions. 

For reference:
RPM packages by group: https://www.rpmfind.net/linux/RPM/Groups.html
Deb packages by section: https://packages.debian.org/stable/

Putting the new tags into an `RPM` block seems like a good solution to avoid confusion.

caarlos0:
Yeah seems good to me! And the deb specific ones we can put under a Deb block too..

willing to make this change?

AdrianWennberg:
Moving the `Section` tag under a Deb block might break current config files if people update nfpm without updating their config. Do you think this will be an issue for people?

Other that that, I'd be happy to make the change!

AdrianWennberg:
Oops forgot to actually create the package in the test.

AdrianWennberg:
Hey Carlos,

I think this is merge ready now. Give me a shout if there is anything else you would like me to add.

rpm/rpm_test.go
```diff
@@ -94,6 +94,16 @@ func TestRPM(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestWithRPMTags(t *testing.T) {
+	var info = exampleInfo()
+	info.RPM = nfpm.RPM{
+		Group:  "default",
+		Prefix: "/usr",
+	}
+	var err = Default.Package(info, ioutil.Discard)
```
caarlos0: maybe write to a file and compare with the expected result? (like goldenfiles above)

AdrianWennberg: I added a new commit. Is that what you had in mind?

caarlos0: looks great, thanks

caarlos0:
thank you very much for the PR and all the effort :) ✨

caarlos0:

PR #90, opened 2019-10-22
fix: remove rpm.Prefix and rpm.Release

caarlos0:
As far as I understand `Prefix` is just to avoid repeating a folder when referencing files.

so, it seems like it didn't make much sense before, and makes less sense now.

IMHO we can just remove it.

`Release` is already a field on the `root` struct, so I think we can remove it as well.


caarlos0:

Issue #55, opened 2019-06-07
add rpm-user


rhd:
Hi, is it possible to either add or workaround fpm's `rpm-user` option?

```
    --rpm-user USER               (rpm only) Set the user to USER in the %files section. Overrides the user when used with use-file-permissions setting.
```

I need the files in the RPM owned by a specific user.

Thanks  and great app!

ti-mo:
Hi @rhd @caarlos0 

Has there been any progress on this? I'm calling nfpm through `goreleaser`, and when installing my rpm in a `fedora:latest` image:

```
[root@66f5aed57836 app]# rpm -Uvh conntracct_0.2.0-next_linux_amd64.rpm
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
Updating / installing...
   1:conntracct-0.2.0-next            
warning: user  does not exist - using root
warning: group  does not exist - using root
################################# [100%]
```

Is this a symptom of this issue? Any known workaround?

caarlos0:
haven't worked on it yet, no...

PRs accepted though :)

ti-mo:
Digging a bit deeper:
```
[root@66f5aed57836 ~]# rpm -qlvp conntracct_0.2.0-next_linux_amd64.rpm
-rw-r--r--    1                                    344 Jan 14 12:59 /etc/conntracct/conntracct.yml
-rwxr-xr-x    1                               14538208 Jan 14 13:16 /usr/bin/conntracct
```

The files inside the RPM don't seem to have an owner at all. Is this intended? I don't need to be able to set the user, but I would just like my package to install without errors/warnings for now. :) In comparison, here is the same command run on another RPM package:

```
[root@66f5aed57836 ~]# rpm -qlvp rpm-5.4.15-42-omv2015.0.x86_64.rpm
-rwxr-xr-x    1 root     root                    32208 Sep 21  2018 /bin/rpm
-rwxr-xr-x    1 root     root                       50 Sep 21  2018 /etc/cron.daily/rpm
```

Could this be a bug?

ti-mo:
Alright, looks like this is indeed being addressed in https://github.com/goreleaser/nfpm/pull/97. Calls to `rpm.AddFile` contain `rpmpack.RPMFile` arguments without `Owner` and `Group` fields set, causing the owner/group fields of all files inside the archive to be empty. RPM does fall back to `root` in this case. With some quick changes applied:

```
		rpm.AddFile(
			rpmpack.RPMFile{
[...]

Issue #15, opened 2018-03-10
Add support for installinit


tonobo:
It would be nice if there where any possibilty to add service files. Debhelper installinit allows to specify several init files which will be automatically managed during install. (e.g enable on install or restart after upgrade). Probaply there is also an option to do the same for rpm.

djmattyg007:
Please be careful about how this is implemented, if at all. The convention on Arch Linux systems is to NOT automatically start services at installation time.

caarlos0:
This will likely not be implemented.

For tips on how to manage init files, check https://nfpm.goreleaser.com/tips/



> Please be careful about how this is implemented, if at all. The convention on Arch Linux systems is to NOT automatically start services at installation time.

we do not support arch just yet...

PR #97, opened 2019-11-03
Added options to add systemd unit and create users

astorath:
changes:
- feat: option to add systemd service unit
- feat: option to add users on package install
- feat: options to define files and directory ownership in config

#55 #15

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -232,63 +232,8 @@ func createControl(instSize int64, md5sums []byte, info *nfpm.Info) (controlTarG
 			return nil, err
 		}
 	}
-
-	scripts := make(map[string]*bytes.Buffer)
-	for _, dest := range []string{"preinst", "postinst", "prerm", "postrm", "rules"} {
-		scripts[dest] = new(bytes.Buffer)
-	}
-
-	if info.SystemdUnit != "" {
-		unit := filepath.Base(info.SystemdUnit)
-		if err := addScriptFromString(scripts["postinst"], "#!/bin/sh\n\nset -e\n\n"); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["prerm"], "#!/bin/sh\n\nset -e\n\n"); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["postrm"], "#!/bin/sh\n\nset -e\n\n"); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["postinst"], strings.ReplaceAll(scriptSystemdPostinst, "#UNITFILE#", unit)); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["prerm"], strings.ReplaceAll(scriptSystemdPrerm, "#UNITFILE#", unit)); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["postrm"], strings.ReplaceAll(scriptSystemdPostrm, "#UNITFILE#", unit)); err != nil {
-			return nil, err
-		}
-	}
-
-	if info.User != "" {
-		if err := addScriptFromString(scripts["preinst"], "#!/bin/sh\n\nset -e\n\n"); err != nil {
-			return nil, err
-		}
-		if err := addScriptFromString(scripts["preinst"], strings.ReplaceAll(scriptCreateUser, "%{package_user}", info.User)); err != nil {
-			return nil, err
-		}
-	}
-
-	for script, dest := range map[string]string{
-		info.Scripts.PreInstall:             "preinst",
-		info.Scripts.PostInstall:            "postinst",
-		info.Scripts.PreRemove:              "prerm",
-		info.Scripts.PostRemove:             "postrm",
-		info.Overridables.Deb.Scripts.Rules: "rules",
-	} {
-		if script != "" {
-			if err := addScriptFromFile(scripts[dest], script); err != nil {
-				return nil, err
-			}
-		}
-	}
-
-	for dest, script := range scripts {
-		if script.Len() > 0 {
-			if err := newScriptInsideTarGz(out, script.Bytes(), dest); err != nil {
-				return nil, err
-			}
-		}
+	if err := addControlScripts(out, info); err != nil {
+		return nil, errors.Wrap(err, "adding contro scripts")
 	}
 
 	if err := out.Close(); err != nil {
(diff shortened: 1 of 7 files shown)


caarlos0:
>     * feat: option to add systemd service unit
> 
>     * feat: option to add users on package install
> 
>     * feat: options to define files and directory users in config

hey, this PR has at least 3 features it seems, can we split them in one PR each?

astorath:
> hey, this PR has at least 3 features it seems, can we split them in one PR each?

@caarlos0 problem is, 
1. they are non-independent
2. they all use the same code for scripting

caarlos0:
you're right.

thinking here: aren't all those things here already possible with the current nfpm version?

is just custom scripts, right? If so, I don't think we should add this to nfpm itself...

astorath:
> thinking here: aren't all those things here already possible with the current nfpm version?
> 
> is just custom scripts, right? If so, I don't think we should add this to nfpm itself...

@caarlos0 

well, consider this:

1. You can't manage files ownership in current version
2. Many (if not most) users don't even know how these scripts look like: rpm as well as deb have a set of helper units providing these functions: https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in, https://packages.debian.org/jessie/dh-systemd, so you never have to write those scripts by hand
3. Scripts are basically the same all the time, no need change anything
4. It's an fpm option already

astorath:
Added references to issues to track for relevance

caarlos0:
> * You can't manage files ownership in current version

this I agree we should add

> * Many (if not most) users don't even know how these scripts look like: rpm as well as deb have a set of helper units providing these functions: https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in, https://packages.debian.org/jessie/dh-systemd, so you never have to write those scripts by hand

if we add systemd, it would make sense to add all the other possible init systems as well, which I rather not do, so that's why maybe its better we don't add any...

> * Scripts are basically the same all the time, no need to change anything

until someone wants to do something different for some reason...

> * It's an fpm option already

this is not fpm, as the name says hehe

caarlos0:
cc/ @djgilcrease any thoughts on this?

deb/deb.go
```diff
@@ -389,3 +486,19 @@ func writeControl(w io.Writer, data controlData) error {
 	})
 	return template.Must(tmpl.Parse(controlTemplate)).Execute(w, data)
 }
+
+func getFilesAttr(raw string) (name, user, mode string) {
```
djgilcrease: this should move somewhere shareable by both rpm and deb

```go
func GetFilesAttr(raw string) (name, user, mode string) {
	var parts []string
        parts = append(parts, strings.SplitN(raw, ":", 3)...)
        parts = append(parts, "", "")
        return parts[0], parts[1], parts[2]
}
```

djgilcrease:
I do not overly mine only supporting systemd, but I see a few issues, if I specify a systemd unit file and already have scripts they would not be merged.

I would prefer a sub command to generate systemd units and associated scripts as actual files so the user can just add them explicitly to the config file or merge it wither their own scripts

caarlos0:
> I would prefer a sub command to generate systemd units and associated scripts as actual files so the user can just add them explicitly to the config file or merge it wither their own scripts

this makes more sense to me as well.

astorath:
@caarlos0  @djgilcrease 
Ok, guys, before I move on to fixing issues, tests etc... Let's fix up what we are doing. To sum up:

- `systemd` is the only service manager on supported versions of RHEL / Amazon Linux 2 /  Debian / Ubuntu, so I don't see any need to support something else (`init`)
- installing program as a linux service is very much vital part of many packages
- `fpm` as well as `rpmbuild` and `dpkg` provides an easy way to do this without much effort or diving into specifics
- adding `systemd` service is not only about hook scripts, but also about adding dependencies, placing service files, etc. - so according to the DRY principle this must be a separate option, not a combination of `bash` stuff on top of CI procedures...

>  if I specify a systemd unit file and already have scripts they would not be merged.
- @djgilcrease actually they would, there are even some tests for the case

ti-mo:
Hi @astorath!

Could I suggest defaulting packaged files to be owned by `root` (at least for RPM)? I've run into some warnings while installing RPMs built by current master, see https://github.com/goreleaser/nfpm/issues/55#issuecomment-574551328 for more details.

Basically, leaving `rpmpack.RPMFile.Owner` and `.Group` empty causes packaged files to have no owner/group:
```
[root@66f5aed57836 ~]# rpm -qlvp conntracct_0.2.0-next_linux_amd64.rpm
-rw-r--r--    1                                    344 Jan 14 12:59 /etc/conntracct/conntracct.yml
-rwxr-xr-x    1                               14538208 Jan 14 13:16 /usr/bin/conntracct
```

as opposed to:
```
[root@689f5577f54f app]# rpm -qlvp conntracct_0.2.0-next_linux_amd64.rpm
-rw-r--r--    1 root     root                      344 Jan 14 12:59 /etc/conntracct/conntracct.yml
-rw-r--r--    1 root     root                      404 Jan 15 07:52 /lib/systemd/system/conntracct.service
-rwxr-xr-x    1 root     root                 14538208 Jan 15 08:15 /usr/bin/conntracct
```

This causes `rpm` to throw warnings during package install:
```
[root@66f5aed57836 app]# rpm -Uvh conntracct_0.2.0-next_linux_amd64.rpm
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
Updating / installing...
   1:conntracct-0.2.0-next            
warning: user  does not exist - using root
warning: group  does not exist - using root
################################# [100%]
```

Patch looks good so far! Adding `info.{User,Group}` also allows user/group support to be implemented in `goreleaser`. :+1:

astorath:
> Could I suggest defaulting packaged files to be owned by root (at least for RPM)? I've run into some warnings while installing RPMs built by current master, see #55 (comment) for more details.

@ti-mo this is not a problem, but I would like see some feedback from team members. @caarlos0 ?

caarlos0:
I think that for now defaulting to root should be good enough yes 🤔

astorath:
> I think that for now defaulting to root should be good enough yes 🤔

Sorry, maybe I should've put it more strait: I need some feedback on all this PR. See my comment on 5 dec.

wg:
Thank you for this contribution @astorath! NFPM is great, and I need support for multiple systemd units which I've implemented here: https://github.com/kentik/nfpm. Once this PR is merged I'll be happy to submit another with my changes.

erikgeiser:
While I appreciate the work and sentiment behind this proposal, I think it is a bad idea in the long run. I think a packaging program should be agnostic about what it is packaging and `systemd` unit files concern the subject that is being packaged rather than the process of packaging. Right now, `nfpm` does not really make assumptions about the target system for which a package is created. Interfacing with programs that may or may not be on the target system can have numerous problems: The interface or API could change, typical workflows could change, the folder structure could change, `systemd` itself could be retired or replaced (probably not gonna happen but who knows). As soon as this starts to happen `nfpm` will either drop support of some target configurations or it will need additional code paths and bloat to accommodate for these changes. At some point `nfpm` would look like `fpm` which is probably undesirable.

Also, in my opinion, magic functionality via post/pre-install scripts is way worse than just embracing new `systemd` features such as [dynamic users](http://0pointer.net/blog/dynamic-users-with-systemd.html) which would be more modern appoach. I would encourage everyone to take a look at all the [powerful features systemd provides](https://www.freedesktop.org/software/systemd/man/systemd.exec.html) and handcraft a unit file that is tailor-made for the program you are packaging an then adding it to the files (or config files) section.

Please correct me if you feel different.

caarlos0:
I think @erikgeiser put it in words better than I could.

That all being said, I think its time to close this.

Thanks everyone! I'm sorry this wasn't merged, but I think its for the best in the long run.

caarlos0:

PR #104, opened 2019-11-11
fix(#103): ensure the package is not included in itself 

### LINKED ISSUE #103: When using `./**/*` for files a 0 byte rpm file gets included in the package
Sample config:
```yaml
name: "app-name"
arch: "noarch"
platform: "linux"
version: "${APP_VERSION}"
files:
  ./**/* : "/some/location/"
```

Assuming I ran the following command: `nfpm pkg -f sample.yaml -t herpderp.rpm`

When the RPM gets installed I would have a 0 byte file by the name of `herpderp.rpm` in `/some/location/`.

I haven't tested if this happens with .deb as well as we only build .rpm packages.

djgilcrease:
ensure the package is not included in itself when using globs to match all files

Closes #103

### DIFF
--- a/rpm/rpm.go
+++ b/rpm/rpm.go
@@ -200,7 +200,7 @@ func createFilesInsideRPM(info *nfpm.Info, rpm *rpmpack.RPM) error {
 				return err
 			}
 			for src, dst := range globbed {
-				if strings.Contains(src, info.Target) {
+				if strings.HasSuffix(src, info.Target) {
 					fmt.Println("skipping", src)
 					continue
 				}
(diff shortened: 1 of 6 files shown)


rpm/rpm.go
```diff
@@ -199,6 +200,10 @@ func createFilesInsideRPM(info *nfpm.Info, rpm *rpmpack.RPM) error {
 				return err
 			}
 			for src, dst := range globbed {
+				if strings.Contains(src, info.Target) {
```
caarlos0: maybe change to `strings.HasSuffix(src, info.Target)`? 

otherwise the target name in the middle of a path would be skipped as well...

regner:
Thanks for doing that, much appreciated!

caarlos0:

Issue #42, opened 2018-11-23
Ability to have a changelog


InTheCloudDan:
I was evaluating nfpm and it would be nice to have the ability to include a changelog file in the generated package. Right now I don't see any way to do it.

djgilcrease:
Would you want this to show up as the rpm changelog or just be a file included in the package?

djgilcrease:
https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog
http://packaging.ubuntu.com/html/debian-dir-overview.html#the-changelog
https://rpm-guide.readthedocs.io/en/latest/rpm-guide.html

Both RPM and Deb have specific formats required for changelogs that differ

**Debian Format**
```
<Package Name> (<Version>-<Release>+<Metadata>) <distribution> ; urgency=low

 * List (Closes: #nnnn)  <nnnn is the bug number of your ITP>
 * of
 * Changes

-- <Packager Name and Email>  <packged date Mon, 22 Mar 2010 00:37:31 +0100>
```

**RPM Format**
```
* <packged date Tue May 31 2016> <Packager Name and Email> - <Version>-<Release>
- List (Closes: #nnnn)
- of
- changes
```

**Data that we already have**
- Version
- Release
- deb-Metadata
- Maintainer
- packaged date


**Data We need**
- DEB: distribution
- DEB: urgency

We will require changelog schema that can be tranformed into both formats. I think the schema (or at least it's auto generation) should be owned by goreleaser as it will help with https://github.com/goreleaser/goreleaser/issues/1131 & https://github.com/goreleaser/goreleaser/issues/990 then nfpm can use that to create the correct changelog data for each format

djgilcrease:
Idea for format that I think will allow people the most flexibility to generate changelog files in the format(s) they wish to support not just `.dep` & `.rpm`

```yaml
- semver: 1.0.0+git.e6c8682
  date: 2006-01-02T15:04:05.999999999Z07:00 // RFC3339Nano
  packager: Carlos A Becker <pkg@carlosbecker.com>
  notes:
    header: |
      New Features
      ============
      There are major new features in this release

      Known Issues
      ============
      bla bla bla
    footer: |
      Header section
      =============
      More things here

      Bla Bla Bla
      =============
      More things here
  changes:
    - commit: e6c8682
      note: Use Version instead of Git tag to for .deb/.rpm
      closes:
        - 1157
    - commit: 50235f2
      note: chore(deps): bump all 
[...]

PR #105, opened 2019-11-15
WIP: feat(#42): Ability to have a changelog

djgilcrease:
**NOTE:** this will NOT create and maintain a changelog for you, you are expected to use https://github.com/goreleaser/chglog to generate and maintain the `changelog.yml` file, this will just read that file in and generate the correct changelog data for deb & rpm

dpkg does not provide any facility to read the changelog of a package. you should extract the package and read the changelog
`dpkg -X <package.deb> <folder>` then you can read the changelog using the dpkg-parsechangelog utility `dpkg-parsechangelog -l <folder>/usr/share/doc/<package>/changelog.Debian.gz`

rpm does provide a way to see the changelog `rpm -q --changelog path/to/pkg.rpm`

**TODO**
* [ ] get https://github.com/goreleaser/chglog to be building and releasing with goreleaser
* [ ] get feedback from @InTheCloudDan & @eugene-babichenko to ensure this meets their expectations


caarlos0:
Overall, it looks good to me.

I only think that the NFPM API should only accept the changelog file and maintainer, and not try to guess anything. We could add those guesses to the cmd though.

I think that because once we integrate chlog with goreleaser, it will already have this info, so we may just pass it down to nfpm...

If running nfpm as a binary, adding this to the cmd should do the trick.

thoughts?

djgilcrease:
Makes sense, will update this after Thanksgiving holiday

erikgeiser:
Changelog support would be essential for me. Is this still in progress? If not, I could maybe look into it and try to fix the issues such that the pull request can be merged soon.

erikgeiser:
I took a look at the pull request, implemented the requested changes, removed dependencies of specific commits of some packages and made it compatible with the master branch. Tomorrow I will take a look at some details such as tests. In a day or two I will submit a new pull request.

caarlos0:
awesome, thanks @erikgeiser

djgilcrease:
superseded by #158

caarlos0:

PR #111, opened 2019-12-05
feat: version release/prerelease options

astorath:
problem to solve:
- rpm and deb have quite different versioning scheme for releases / prereleases:
  https://fedoraproject.org/wiki/Package_Versioning_Examples#Complex_versioning_examples
  https://www.debian.org/doc/manuals/maint-guide/first.en.html#namever
- it is impossible right now to use `release` version for nfpm deb package (because `release` part is actually used as prerelease)
 
changes:

- added `prerelease` option to specify it separately from `release`
- for deb: `prerelease` comes after tilde, `release` after dash (breaking!)
- for rpm: `prerelease` comes with default `0.1` release, `release` is not changed
- semver parsing now treats numeric `prerelease` data as release version
- semver parsing now treats non-numeric `prerelease` data as prerelease version


nfpm.go
```diff
@@ -48,6 +50,18 @@ func Parse(in io.Reader) (config Config, err error) {
 	config.Info.Release = os.ExpandEnv(config.Info.Release)
 	config.Info.Version = os.ExpandEnv(config.Info.Version)
 
+	// parse the version as a semver so we can properly split the parts and support proper ordering for both rpm and deb
```
caarlos0: this entire block should be in the `withdefaults` method, as when using goreleaser as a lib this method is not used.

nfpm.go
```diff
@@ -48,6 +50,18 @@ func Parse(in io.Reader) (config Config, err error) {
...
+	var v *semver.Version
+	if v, err = semver.NewVersion(config.Info.Version); err == nil {
+		config.Info.Version = fmt.Sprintf("%d.%d.%d", v.Major(), v.Minor(), v.Patch())
+		if config.Info.Release == "" && valid.IsInt(v.Prerelease()) {
```
caarlos0: I would prefer to not import an entire lib just for int validation, could add a new func:

```go
func isInt(s string) bool {
  _, err := strconv.Atoi(v)
  return err == nil
}
```

and then:

```suggestion
		if config.Info.Release == "" && isInt(v.Prerelease()) {
```

rpm/rpm.go
```diff
@@ -104,17 +104,24 @@ func buildRPMMeta(info *nfpm.Info) (*rpmpack.RPMMetaData, error) {
 		return nil, err
 	}
 
-	hostname, err := os.Hostname()
+  hostname, err := os.Hostname()
```
caarlos0: format...

rpm/rpm.go
```diff
@@ -104,17 +104,24 @@ func buildRPMMeta(info *nfpm.Info) (*rpmpack.RPMMetaData, error) {
...
 		return nil, err
 	}
 
+	var release = ""
+	if info.Prerelease == "" {
+		release = defaultTo(info.Release, "1")
+	} else {
+		release = fmt.Sprintf("%s.%s", defaultTo(info.Release, "0.1"), info.Prerelease)
+  }
+
```
caarlos0: this could be:

```go
var release = defaultTo(info.Release, "1")
if info.PreRelease != "" {
  release = fmt.Sprintf("%s.%s", defaultTo(info.Release, "0.1"), info.Prerelease)
}
```

nfpm.go
astorath: ok

nfpm.go
astorath: Got it, sry

caarlos0:
thanks, sorry for  the delay

caarlos0:

PR #113, opened 2019-12-16
fix: updated architectures for arm5 and arm6

djthorpe:
I updated your ARM architecture as ARM6 is actually armhf and ARM5 is armel (I think) as per this page: https://github.com/golang/go/wiki/GoArm


caarlos0:
That is correct, thanks!

caarlos0:

PR #119, opened 2020-02-18
fix: Handle multiline input for the Debian description field

djboris9:
Hi There

The description field in the DEB control file supports multiline/verbatim input. It is done by indenting the lines by two spaces. Before this patch, the control file was broken if a user wanted to supply a multiline description.

Blank lines by using a full-stop-character are not implemented as it requires to convert the whole block to a 1-level intended block instead of the verbatim 2-level intended block.

Ref: [debian-policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#description)


caarlos0:
thanks for the PR :)

caarlos0:

PR #123, opened 2020-03-13
Fix glob patterns for files and config_files

matlec:
We use nfpm for creating Debian packages. In our nfpm.yaml file we make use of glob patterns for `files`. In the following example we copy all artifacts from `a` to `/tmp/a` and from `b` to `/tmp/b`:
```
files:
  "./a/**/*": "/tmp/a/"
  "./b/**/*": "/tmp/b/"
```

nfpm does not work as expected when the directories `a` or `b` only contain a single file (e.g. ./a/test.txt). In this case nfpm creates a deb package that contains the content of ./a/test.txt as a file at the path /tmp/a. Instead we expect ./a/test.txt to be packaged as file at the path /tmp/a/test.txt.

For our use cases this fix is kind of important and the only work-around right now is to make sure that there are always at least two files below a and b.

### DIFF
--- a/glob/glob.go
+++ b/glob/glob.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/mattn/go-zglob"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


matlec:
It would be nice if you could push a new version that contains this fix :-)

caarlos0:
thanks for the PR 🚀

caarlos0:
> It would be nice if you could push a new version that contains this fix :-)

releasing...

matlec:
Thank you, works like a charm

caarlos0:

Issue #39, opened 2018-10-25
Add support for apk


tcurdt:
Would be nice to have support for alpine linux packages as well.

http://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package
https://wiki.alpinelinux.org/wiki/Alpine_package_format

caarlos0:
interested in try to implement a PR for this?

tcurdt:
I am still investigating using docker and the native tools - but consider me tempted :)
I just don't get why all those native packaging tools are so incredibly bad.

hellozee:
> cat $controldir/control.tar.gz $controldir/data.tar.gz > mypackage-1.0-r0.apk

If that's the only thing to be done, I could give it a shot

tcurdt:
I am currently working on it and in the end (as always) it is a little bit more involved.
But the prototype is almost there. An unsigned `apk` already installed successfully.
Currently working on the signing.

tcurdt:
@hellozee maybe take a stab at native `rpm` support instead? #40

hellozee:
@tcurdt cool, but well `rpm` is something, I have never taken a look, I just used `alien` to convert `debs` to `rpm` if I needed any

tcurdt:
Got the `apk` signing working as well 🎉🎉🎉
Some of the hashing is still a bit odd (as the host system does not seem to care if it's there or not) - but the package now installs even in a trusted manner.

tcurdt:
Still need to find the time to wrap this up in a PR.

chris-rock:
@tcurdt that would be an awesome addition. Any progress?

tcurdt:
@chris-rock I got it working. It's just sitting on my machine waiting until I find the time to turn it into a proper PR. Distracted by other projects - but I'll try to make some time.

tcurdt:
Just a quick status update: I've started to massage the code base and tests to accept the new apk packager. Now I need to migrate the code into the packager.

chris-rock:
@tcurdt I would love to try your implementation. Let me know when its ready for a beta test 🤣

tcurdt:
@chris-rock TBH I am not sure when I will find the time to finish this :-/
If you want I could link a gist with the experimental code - if that helps?

chris-rock:
I would appreciate if you could make your current state available. Then the community can collaborate to implement the feature.

tcurdt:
[...]

PR #126, opened 2020-03-27
feat: Add apk support

bhamail:
Work in Progress - Not ready to commit, just looking for guidance

This is my first stab at getting Torsten Curdt's work from Issue #39 into a PR to add generation of .apk installers.

As you can see, I wander without much of a clue, so any pointers you can give as to "what to do next" are most welcome!

Most of the linter issues are fixed, but the big nasty `runit` method (formerly known as `main`)  still needs refactoring...


caarlos0:
sorry, I haven't had time to look at this just yet... but I promise I will :)

caarlos0:
(I did take a quick look on it though, and its looking good so far. Linting issues we can solve later :) )

bhamail:
@caarlos0 I just noticed your comments. Thanks so much for looking. I'm really running with very little clues in hand (as you can now see). :) Willing to learn!

caarlos0:
no problem @bhamail , let me know if you need anything :)

bhamail:
Any suggestion you can give about how to get this PR really working would be great. Right now, I've only taken the original authors work and put it into the project. However, it doesn't actually do anything right now. I think it should probably re-use the .deb package code to build (or reuse) .deb installer parts, but not even sure where to start.

caarlos0:
you pretty much just need to implement this interface https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L68

and then register the format: https://github.com/goreleaser/nfpm/blob/master/rpm/rpm.go#L21-L24

For now if you need similar code, just copy it, we can make a shared lib or something later if its worth it :)

bhamail:
@caarlos0 Sorry to be dense, but my brain is starting to go in circles.

I started down the path of rewriting the `createGzData()` method to take a `info *nfpm.Info` parameter instead of `dataTgz io.Writer`. The idea being to use values in `nfpm.Info` instead of a tree on disk. 

That quickly felt like the wrong way to go because it was so similar to (but different from) logic in deb.go. I started to duplicate the [deb.go -> createDataTarGz()](https://github.com/goreleaser/nfpm/blob/master/deb/deb.go#L92) logic of:

```
// handle empty folders

// handle Files and ConfigFiles
```

inside [apk.go -> createBuilderData()](https://github.com/goreleaser/nfpm/pull/126/files#diff-80a1dad5fd9269e49adca48a5179860cR378).

I think I'm losing site of the forest for the trees, and would appreciate any pointers on how to proceed. Maybe @tcurdt could steer me back on course too?

At a high level, I thought I should be able to copy the .deb logic, and add some additional signing, and I think I'm missing an easier path forward.

tcurdt:
Oh, I wasn't aware someone is working on this. Great! I am happy to help as good as I can. I will have a look tomorrow. Do you have any questions in particular, @bhamail ?

tcurdt:
I am seeing the following sizes of the temp files
```
-rw-r--r--  1 tcurdt  staff  1346 Apr  6 23:54 apkToCreate.apk
-rw-r--r--  1 tcurdt  staff   302 Apr  6 23:54 apk_control.tgz
-rw-r--r--  1 tcurdt  staff   378 Apr  6 23:54 apk_data.tgz
-rw-r--r--  1 tcurdt  staff   666 Apr  6 23:54 apk_signatures.tgz
```
Where did you get those files sizes from for the verifcation?

apk/apk_test.go
```diff
@@ -0,0 +1,116 @@
...
+	verifyFileSize(t, path.Join(tempDir, "apk_control.tgz"), 304, 300, 305)
+	verifyFileSize(t, path.Join(tempDir, "apk_data.tgz"), 414, 373, 407)
+	verifyFileSize(t, path.Join(tempDir, "apk_signatures.tgz"), 665, 665, 667)
+}
+
+func verifyFileSize(t *testing.T, fileToVerify string, expectedSize, expectedSizeCiMin, expectedSizeCiMax int64) {
+	fi, err := os.Stat(fileToVerify)
+	assert.Nil(t, err)
+	ciEnv := os.Getenv("CI")
+	if ciEnv != "" {
```
tcurdt, line 46: Why is it that the CI would need a size range?

apk/testdata/apkTest.dockerfile
```diff
@@ -0,0 +1,25 @@
+# run using:
+#docker build -t dantest -f ./apk/testdata/apkTest.dockerfile . && docker run -it dantest
+
+#FROM alpine
+FROM golang:1.13-alpine
+
+WORKDIR /go/src/app
+COPY . .
```
tcurdt, line 8: move the COPY further down, right before running the compile/test, that will be much quicker during repeated testing as it allows to cache more of the layers.

apk/apk.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		return err
+	}
+	defer dataTgz.Close()
+
+	size := int64(0)
+	sizep := &size
+
+	// create the data tgz
+	dataDigest, err := createGzData(dataTgz, pathToFiles, sizep, size)
```
tcurdt, line 250: I'd consider the `gz` an implementation detail, I would just call it `createData`

apk/apk.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		return err
+	}
+
+	// create the control tgz
+	builderControl := createBuilderControl(size, dataDigest, err)
```
tcurdt, line 256: Maybe use the same pattern as for the `data` section using `createControl` function?

apk/apk.go
```diff
@@ -0,0 +1,424 @@
...
+		return err
+	}
+	signed, err := priv.Sign(rand.Reader, controlDigest, crypto.SHA256)
+	if err != nil {
+		return err
+	}
+	fmt.Println("data sign  :", hex.EncodeToString(signed))
+
+	// create the signature tgz
+	builderSignature := createBuilderSignature(signed, err)
```
tcurdt, line 280: Maybe use the same pattern as for the `data` section using `createSignature` function?

apk/apk.go
```diff
@@ -0,0 +1,424 @@
...
+	_, err = writeTgz(signatureTgz, tarCut, builderSignature, sha256.New())
+
+	// combine
+	return combineGzsToFile(target, signatureTgz, controlTgz, dataTgz)
```
tcurdt, line 284: I'd either name it `combineTgzsTo()` or just `combineToApk()`

apk/apk.go
tcurdt: Having the trinity of `createData`, `createControl` and `createSignature` will make the code easier to read and easier to follow the file format IMO.

tcurdt:
I didn't get around to test the generated `apk` yet - but this looks promising!

tcurdt:
I am sure there is overlap with the `deb` part - but the support of `full/cut` and the alignment in `writeTgz` is crucial to get this working for `apk`. But you are right - looking at the `deb` I guess it should use `*nfpm.Info` instead of straight walking. That said, I think I'd focus first on getting the test working and having an `apk` that installs successfully. Next it would probably make sense to refactor and decouple the `tgz` code that both `deb` and `apk` then can use. Just my 2 cents.

bhamail:
> I am seeing the following sizes of the temp files
> 
> ```
> -rw-r--r--  1 tcurdt  staff  1346 Apr  6 23:54 apkToCreate.apk
> -rw-r--r--  1 tcurdt  staff   302 Apr  6 23:54 apk_control.tgz
> -rw-r--r--  1 tcurdt  staff   378 Apr  6 23:54 apk_data.tgz
> -rw-r--r--  1 tcurdt  staff   666 Apr  6 23:54 apk_signatures.tgz
> ```
> 
> Where did you get those files sizes from for the verifcation?

The "non-ci" files sizes is what is created on my machine (mac os).
The "ci" file size ranges are what I saw getting creating during ci run. 
Yes, they should not differ, but this code as is now is not fully functional. I just slapped something in there that "passes" ci - and run locally. 

What is the code actually expecting to find in the `pathToFiles` location? I couldn't tell for sure, so not even sure how to flesh out the unit tests.

apk/testdata/apkTest.dockerfile
bhamail: > ...move the COPY...

Cool! Done in b190238

BTW, I don't actually use this file yet in an automated way. I was manually stopping the unit test that created an .apk, copying the .apk somewhere to save it, then spinning up the docker image and trying to install that .apk.

apk/apk_test.go
bhamail: > Why is it that the CI would need a size range?

Only because stuff is broken. :) This whole PR is not ready for prime time, I'm just trying to get the ball rolling forwards.

apk/apk.go
bhamail: > I'd consider the `gz` an implementation detail, I would just call it `createData`

Done b63d79d

apk/apk.go
bhamail: > Maybe use the same pattern as for the `data` section using `createSignature` function?

Agreed, but I wanted to have a more valid unit test suite of a working .apk, so I could be confident I was actually "refactoring", and not just changing stuff hoping for the best.

I also expect I will need to change the "private key" logic to take base64 encoded string (typically provided by an environment variable in many CI systems), rather than a path to a private key file on disk. So I was holding off on the stuff until I better understood what the createData() and createControl() code is actually doing.

apk/apk.go
bhamail: > I'd either name it `combineTgzsTo()` or just `combineToApk()`

Done in 9d5bbd8

tcurdt:
> What is the code actually expecting to find in the pathToFiles location?

That's the path to the files in the apk. What goes in there depends only on the package maintainer.
Since my first implementation was running outside of `nfpm` it needed a path where to find the files to add.

Have a look at #39. This is the equivalent how the `dataTgz` is being built:

```
cd $pkgdir
tar -c * | abuild-tar --hash | gzip -9 > $controldir/data.tar.gz
```

The `controlTgz` holds all the metadata of the package.
The `signatureTgz` holds the package signature.

After some modifications of the docker file and some more digging I got the `apk` generated by the test to actually install:

```
# mkdir -p /testdata/files
# apk add -vvv --allow-untrusted apkToCreate.apk 
The following NEW packages will be installed:
  xbase
After this operation, 0 KiB of additional disk space will be used.
(1/1) Installing xbase (1.0.0-r1)
testdata/files (dir)
testdata/files/control.golden
OK: 32 packages, 161 dirs, 1324 files, 117 MiB
```

To me it seems it's `createBuilderData` needs to be moved to use files provided by `nfpm` instead of walking to finish the first step. This will also take care of the `mkdir` I had to use.

Next will be to use the metadata from `nfpm` inside `createBuilderControl`.

I guess testing will need some work. But those should be the major parts to finish this up.

tcurdt:
I didn't want to do the whole PR thing for the little changes to get it working:

[changes.diff.zip](https://github.com/goreleaser/nfpm/files/4443249/changes.diff.zip)

bhamail:
> I didn't want to do the whole PR thing for the little changes to get it working:
> 
> [changes.diff.zip](https://github.com/goreleaser/nfpm/files/4443249/changes.diff.zip)

@tcurdt Thanks for the help with this! I just added a `skipVerify` flag to avoid the file size dance while testing.

I was scratching my head about why the .apk install was failing. Thanks for the `mkdir` workaround, which makes it clearer why the install was failing.

I will dive into rewriting the `createBuilderData()` method to take a `info *nfpm.Info` next.

tcurdt:
The size verification after compression is probably isn't the best way to verify the test results - but I am still a little surprised there actually is a difference.

bhamail:
> The size verification after compression is probably isn't the best way to verify the test results - but I am still a little surprised there actually is a difference.

I totally agree. Wanted "something" to verify I didn't break your original code as I refactored it into smaller methods to get CI happy. (e.g. the local size tests is how I realized I had to pass a size pointer instead). I intend to replace all my "verify size" cruft with "real" tests as I go. The end goal being no more goofy size woobly, wibly range madness.

bhamail:
@tcurdt  I changed `createData()` to use `info *nfpm.Info`. Please holler about any atrocities you see.

I think the next step is "use the metadata from nfpm inside `createBuilderControl()`". 
@caarlos0  Any pointers on what to lift from `deb.go` for this?

tcurdt:
I didn't check things in detail but overall it looks like good progress. Nice.

For CI you definitely want support for via env var. Might be nice to also have a way to use a file when running outside of CI. But I guess should align with the general approach of `nfpm`.

bhamail:
> I didn't check things in detail but overall it looks like good progress. Nice.
> 
> For CI you definitely want support for via env var. Might be nice to also have a way to use a file when running outside of CI. But I guess should align with the general approach of `nfpm`.

Cool. I included a helper method (`apk.go -> fileToBase64String(file string) (string, error)`) that reads a private key file from disk into a base64 encoded string for use in the packager. Hopefully that will suffice for running outside CI.

bhamail:
@caarlos0 (or anyone who knows). How do I trigger the `apk.go-init()` method?
Is there some other step needed to ensure the format is registered?

I'm doing some local testing, and from the cli, I keep getting the error below (due to non-registered apk format):

```
nfpm pkg --target fake.apk
nfpm: error: no packager registered for the format apk
```

tcurdt:
> @caarlos0 (or anyone who knows). How do I trigger the `apk.go-init()` method?
> Is there some other step needed to ensure the format is registered?

Pretty sure you need to add

```
	_ "github.com/goreleaser/nfpm/apk"
```

to `main.go` for the init to be called.

bhamail:
> > @caarlos0 (or anyone who knows). How do I trigger the `apk.go-init()` method?
> > Is there some other step needed to ensure the format is registered?
> 
> Pretty sure you need to add
> 
> ```
> 	_ "github.com/goreleaser/nfpm/apk"
> ```
> 
> to `main.go` for the init to be called.

Bam! Nifty. Thanks.

bhamail:
Yaml question: I am getting the error below using the config shown. I assume I've botched either the yaml format for the config file, and/or the [Overridables](https://github.com/goreleaser/nfpm/pull/126/files#diff-10fa0544ff0e910747a6ec13e04cea29R142) or [Apk](https://github.com/goreleaser/nfpm/pull/126/files#diff-10fa0544ff0e910747a6ec13e04cea29R163) struct. Anyone see what I am missing?

```
nfpm pkg --target fake.apk
nfpm: error: yaml: unmarshal errors:
  line 47: field privatekey not found in type nfpm.Overridables 
```

config file: (I realize the `privatekey` test value is invalid, but this fails sooner, during yaml parsing I think).
```
conflicts:
- not-foo
- not-bar
maintainer: "John Doe <john@example.com>"
description: |
  FooBar is the great foo and bar software.
    And this can be in multiple lines!
vendor: "FooBarCorp"
homepage: "http://example.com"
license: "MIT"
bindir: "/usr/local/bin"
files:
  ./foo: "/usr/local/bin/foo"
  ./bar: "/usr/local/bin/bar"
config_files:
  ./foobar.conf: "/etc/foobar.conf"
overrides:
  rpm:
    scripts:
      preinstall: ./scripts/preinstall.sh
      postremove: ./scripts/postremove.sh
  deb:
    scripts:
      postinstall: ./scripts/postinstall.sh
      preremove: ./scripts/preremove.sh
  apk:
    privatekey: "yaddayadda"
```

tcurdt:
It fails during unmarshalling, not parsing.

It tries to unmarshall the key to https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L129 which does not have such a key. It's missing the reference to `apk`. But I am not seeing an obvious cause either.

bhamail:
> It fails during unmarshalling, not parsing.
> 
> It tries to unmarshall the key to https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L129 which does not have such a key. It's missing the reference to `apk`. But I am not seeing an obvious cause either.

@caarlos0 I'm not having any luck either. I've added the [Apk](https://github.com/goreleaser/nfpm/pull/126/files#diff-10fa0544ff0e910747a6ec13e04cea29R142) struct to nfpm->Overridables. 
I did make sure to rebuild the `nfpm` executable.
Any ideas?

FYI, I added a `PrivateKeyFile` field to that struct. `PrivateKey` will take precedence if configured.

caarlos0:
the overridables thing is supposed to be for, for example, override a field that's common on all formats for a single format.

other formats don't have this private key field, so I would guess `mergo` (lib that does the merge) fails because of that?

I think we can probably remove it to the root nfpm struct instead of overridables...

bhamail:
> the overridables thing is supposed to be for, for example, override a field that's common on all formats for a single format.
> 
> other formats don't have this private key field, so I would guess `mergo` (lib that does the merge) fails because of that?
> 
> I think we can probably remove it to the root nfpm struct instead of overridables...

Aha! Thanks for you explanation @caarlos0. I moved the PrivateKey* configs to root, and also read them from env vars if configured.

Now I can actually run `nfpm` and generate an `.apk` that installs. Feels like it's getting close.

Suggestions on what should happen next?

apk/apk.go
```diff
@@ -0,0 +1,597 @@
...
+			return err
+		}
+	}
+
+	return err
+}
+
+func createBuilderSignature(signed []byte, err error) func(tw *tar.Writer) error {
+	return func(tw *tar.Writer) error {
+		keyname := "alpine-devel@lists.alpinelinux.org-4a6a0840"
```
tcurdt, line 311: This should probably be configurable.

bhamail: > This should probably be configurable.

Agreed. Added a new setting at the root.  769b4ad

bhamail:
I could use some help testing the 'scripts' support, as I'm getting errors using the test scripts. Here's an example of error I'm seeing:

```
The following NEW packages will be installed:
  nancy
After this operation, 11 MiB of additional disk space will be used.
(1/1) Installing nancy (1.0.1)
Executing nancy-1.0.1.pre-install
ERROR: nancy-1.0.1.pre-install: script exited with error 127
usr/ (dir)
usr/local/ (dir)
usr/local/bin/ (dir)
usr/local/bin/nancy
Executing nancy-1.0.1.post-install
ERROR: nancy-1.0.1.post-install: script exited with error 127
1 error; 28 packages, 143 dirs, 1270 files, 124 MiB
The command '/bin/sh -c apk add -vvv --allow-untrusted ./dantest/nancy.apk' returned a non-zero code: 1
```

Here's what exists in the .apk:
```
/go/src/app # tar -xvf dantest/nancy.apk 
.SIGN.RSA..rsa.pub
.PKGINFO
.pre-install
.post-install
.pre-deinstall
.post-deinstall
usr/
usr/local/
usr/local/bin/
usr/local/bin/nancy
```

Maybe an issue in how the script files are named in the .apk? (I'm stumped)
 
Also, I'm not even sure I'm reading the right "specs" for the .apk layout. Here are a couple pages I've been reading. Any better suggestions?
https://wiki.adelielinux.org/wiki/APK_internals
https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package#Configure_options

Thanks!

tcurdt:
To me it reads as if it tries to execute `nancy-v.post-install` which is not part of package. Not sure if that's filename or just a reference to the package.

Things to check:
1. What's the content of the script
2. What are the permissions of the script
3. How do scripts look like in official packages

tcurdt:
Specs and information should be linked in the original issue.

bhamail:
> Specs and information should be linked in the original issue.

I was a bit concerned about the second link:
Screen Shot 2020-04-10 at 5 45 10 PM

My google foo yielded some newer links, but wasn't sure what was authoritative and/or current.

I used the same scripts from the nfpm tree - very simple, and changes to the script content don't appear to have any effect. It feels like the script is not actually being executed, and I can't figure out how to get more debug info at installer runtime.

The first few .apk files I grabbed of the web didn't seem to extract with `tar`. Could you point me to a good .apk example that I can disect?

tcurdt:
I will have to check tomorrow - but IIRC I was able to dis-tangle the parts of the `apk` before quite easily. Did you try `apk info -a` and `apk info -L`? My gut still tells me there is a filename mismatch.

tcurdt:
Run an alpine image
```
docker run -it --rm alpine  
```

Then download an official package:
```
# curl -L http://dl-cdn.alpinelinux.org/alpine/v3.8/main/x86_64/nginx-1.14.2-r2.apk > foo.apk
# tar xzvf foo.tgz 
.SIGN.RSA.alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub
.PKGINFO
.pre-install
.pre-upgrade
.post-upgrade
etc/
etc/logrotate.d/
etc/logrotate.d/nginx
etc/init.d/
etc/init.d/nginx
etc/nginx/
etc/nginx/scgi_params
etc/nginx/fastcgi_params
etc/nginx/uwsgi_params
etc/nginx/fastcgi.conf
etc/nginx/mime.types
etc/nginx/conf.d/
etc/nginx/conf.d/default.conf
etc/nginx/modules/
etc/nginx/nginx.conf
usr/
usr/sbin/
usr/sbin/nginx
usr/lib/
usr/lib/perl5/
usr/lib/perl5/vendor_perl/
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/auto/
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/auto/nginx/
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/auto/nginx/nginx.so
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/auto/nginx/.packlist
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/nginx.pm
usr/lib/perl5/vendor_perl/x86_64-linux-thread-multi/perllocal.pod
usr/lib/perl5/vendor_perl/man3/
usr/lib/perl5/vendor_perl/man3/nginx.3pm
usr/lib/nginx/
usr/lib/nginx/modules/
var/
var/tmp/
var/tmp/nginx/
var/log/
var/log/nginx/
var/lib/
var/lib/nginx/
var/lib/nginx/tmp
var/lib/nginx/html/
var/lib/nginx/html/index.html
var/lib/nginx/html/50x.html
var/lib/nginx/modules
var/lib/nginx/run
var/lib/nginx/logs
var/www/
var/www/localhost/
var/www/localhost/htdocs/
```

Next step would be to compare.

bhamail:
> ... Then download an official package:

@tcurdt Thanks for you patient guidance. It looks like a "shell thang", where the scripts I had didn't follow the pattern of:
```bash
#!/bin/sh

do some stuff

exit 0
```

I thought I had tried all combinations of adding "shebang" and "exit 0" to my test scripts, but clearly I did not. 
I think this means the script support actually works!

```
# apk add -vvv --allow-untrusted ./dantest/nancy.apk
The following NEW packages will be installed:
  nancy
After this operation, 11 MiB of additional disk space will be used.
(1/1) Installing nancy (v)
Executing nancy-v.pre-install
running preinstall.sh
usr/ (dir)
usr/local/ (dir)
usr/local/bin/ (dir)
usr/local/bin/nancy
Executing nancy-v.post-install
running postinstall.sh
OK: 28 packages, 143 dirs, 1270 files, 124 MiB
```

Thoughts on what to do next? (I think I'm getting perilously close the deleting my most offensive tgz size based test soon).

tcurdt:
Awesome - good find.

Seems now it's mostly testing left. I guess I would untargz the `apk` and then check for shasums and meta data. But I guess this is where @caarlos0 has to chime in and declare what he thinks is left to be done for a merge.

caarlos0:
Hey!

Thanks for the great work y'all :)

I think the only thing missing is integration tests.. there is a folder with a bunch of them, we can pretty much just add the ones for apk and that should be it AFAIK 🤔

bhamail:
> I think the only thing missing is integration tests.. there is a folder with a bunch of them, we can pretty much just add the ones for apk and that should be it AFAIK

Looking into `acceptance_test.go`, I figured out how to add in `apk` to the `formats` array (and the nifty `_ "github.com/goreleaser/nfpm/apk"` import too). However, I'm not sure how to add a specific config for the 'apk' config. Apk format requires `nfpm.Info.PrivateKeyBase64` or `PrivateKeyFile` (for signing the package).

tcurdt:
@bhamail IIRC `apk` signing is optional and I added that later on. Maybe that's something that should also be made optional. That said - in the real world you want it. FWIW I guess that does not make the integration tests easier. At best it adds another path to test. Just wanted to mention that.

nfpm.go
```diff
@@ -106,23 +109,26 @@ func (c *Config) Validate() error {
...
+	Prerelease       string `yaml:"prerelease,omitempty"`
+	Section          string `yaml:"section,omitempty"`
+	Priority         string `yaml:"priority,omitempty"`
+	Maintainer       string `yaml:"maintainer,omitempty"`
+	Description      string `yaml:"description,omitempty"`
+	Vendor           string `yaml:"vendor,omitempty"`
+	Homepage         string `yaml:"homepage,omitempty"`
+	License          string `yaml:"license,omitempty"`
+	Bindir           string `yaml:"bindir,omitempty"`
+	PrivateKeyBase64 string `yaml:"privatekeybase64,omitempty"`
```
caarlos0, line 128: for those `apk`-only properties, maybe we could create a `Apk` struct?

check for instance `RPM` and `Deb` bellow in this same file.

caarlos0:
> Is it OK just to add the key config to `simple.yml`, and others? I didn't want to pollute all the configs if there was a more specific way to add it.

yes, just note my comment above about the apk struct...

tcurdt:
> to be fair nfpm does not sign the other formats either, so maybe we could even leave this out for the future 🤔

But given it's working already... ?

Maybe time to also add it to for the other packages instead?
Unsigned packages is a bit like online banking over http :)

It's been a while but at least for `deb` I might be able to provide the very least [some guidance](https://github.com/tcurdt/jdeb/blob/master/src/main/java/org/vafer/jdeb/DebMaker.java#L474).

bhamail:
> But given it's working already... ?
> 
> Maybe time to also add it to for the other packages instead?

I'm game to try adding signing to other packagers if others are.
That said, it might make sense to do it incrementally, which would make signing optional - which means I should probably also make signing optional for apk? (I think I just talked myself into   circle).

There's one thing I'd like to solve first: I still can't install .apk's unless I have the `--allow-untrusted` parameter, which I think means there is still some issue with how these are signed and/or how I've put the .pub key into the OS (see the various apk dockerfile IT's where I copy the public key to `/etc/apk/keys/appuser@3b1a5e094ee4.rsa.pub`, e.g.: https://github.com/goreleaser/nfpm/pull/126/files#diff-772ef48e7f3a457cc414244fbca88e34R4). With any luck, I'm just missing something simple - but I don't want to claim it's fully working just yet.

tcurdt:
> That said, it might make sense to do it incrementally, which would make signing optional - which means I should probably also make signing optional for apk?

Of course incrementally. I was just suggesting to make it optionally and not remove it.

>  I still can't install .apk's unless I have the --allow-untrusted parameter

That's just because your OS doesn't know about the key yet. That's completely expected. Just because a package is signed doesn't mean it is trusted. I had this working.

> For this example, we will use alpine-devel@example.com-5629d7e6.rsa which will create the alpine-devel@example.com-5629d7e6.rsa and alpine-devel@example.com-5629d7e6.rsa.pub keypair and copy the public key to /etc/apk/keys/. [...] For hosts to trust this index ensure that the public component of the key you generated exists in their /etc/apk/keys directory. 

IIRC the filename of the key was important.

caarlos0:
> But given it's working already... ?
> 
> Maybe time to also add it to for the other packages instead?
> Unsigned packages is a bit like online banking over http :)
> 
> It's been a while but at least for `deb` I might be able to provide the very least [some guidance](https://github.com/tcurdt/jdeb/blob/master/src/main/java/org/vafer/jdeb/DebMaker.java#L474).

I was thinking more in the sense of avoiding breaking changes to the config later, maybe use same fields for all formats and etc...



> I'm game to try adding signing to other packagers if others are.

awesome! much appreciated 🙏





> That said, it might make sense to do it incrementally, which would make signing optional - which means I should probably also make signing optional for apk? (I think I just talked myself into circle).

yeah, I think it makes sense, only sign if all signing settings were provided or something like that...

caarlos0:
So I guess what I'm trying to say is, maybe: 

- remove the signing from this PR 
- make more PRs adding signing to all formats, trying to use the same configs for all

does that make sense? 🤔

bhamail:
> > I still can't install .apk's unless I have the --allow-untrusted parameter
> 
> That's just because your OS doesn't know about the key yet. That's completely expected. Just because a package is signed doesn't mean it is trusted. I had this working.
> 
> > For this example, we will use [alpine-devel@example.com-5629d7e6.rsa](mailto:alpine-devel@example.com-5629d7e6.rsa) which will create the [alpine-devel@example.com-5629d7e6.rsa](mailto:alpine-devel@example.com-5629d7e6.rsa) and [alpine-devel@example.com-5629d7e6.rsa.pub](mailto:alpine-devel@example.com-5629d7e6.rsa.pub) keypair and copy the public key to /etc/apk/keys/. [...] For hosts to trust this index ensure that the public component of the key you generated exists in their /etc/apk/keys directory.
> 
> IIRC the filename of the key was important.

Here is an example of an IT in this PR where the public key is copied into the OS:
https://github.com/goreleaser/nfpm/pull/126/files#diff-772ef48e7f3a457cc414244fbca88e34R4

The name of the file matches the email used to create the key. Is there more needed to let the OS "know" about the key?

tcurdt:
On the first glance it looks fine to me. But it's been too long since I played with that. I only remember that matching the filename up correctly was important. Again I would suggest to compare to an official package to get it right. Maybe also check the file ownership and permissions. If they are not correct it could be that the key is being ignored - even if it **has** the correct name.

tcurdt:
But it should not be more than the filename+ownership+permissions and keyname in the metadata that needs to align.

tcurdt:
Any update on this? Would be great to have this PR not go stale :)

caarlos0:
waiting on replies to: https://github.com/goreleaser/nfpm/pull/126#issuecomment-614624822

tcurdt:
Would it be OK to just comment out the signing? Or do you want it to be removed removed?

caarlos0:
> Would it be OK to just comment out the signing? Or do you want it to be removed removed?

can be commented, yes, then we can work on adding it to all packages :)

tcurdt:
@bhamail Can you find the time? Or should someone else take over? Would be a shame to have the PR to rot away :)

Slach:
;(( oh, so nice feature, I'm waiting for it so long ;)
@bhamail  could you get me `write` access to your github fork, maybe I can continue this work?

DarthHater:
Hi! I'm the person who works with @bhamail at Sonatype, I'll see if we can carve out some time to finish this up!

bhamail:
Would be happy to have some help on this. New eyes would likely see what I'm missing.

bhamail:
@Slach Please feel free to fork from my fork. You can always do a PR back to mine.

caarlos0:
check #207 please :)

tcurdt:
I did not try it - but looking at the diff it looks OK to me. Cool, there is progress :)

caarlos0:
if you folks can take a look at https://github.com/goreleaser/nfpm/pull/207 

build passing, merges done, etc :) 

thanks!

bhamail:
> if you folks can take a look at #207
> 
> build passing, merges done, etc :)
> 
> thanks!

LGTM! Left a comment on that PR. Huzzah!

caarlos0:

PR #136, opened 2020-05-13
refactor: add a method on Info to list files to copy to the package

dzhu:
Since iterating over the files to copy to the package is something that
all package formats need, adding a central place for that logic reduces
duplication.

Also sort the list of files as a step toward making builds more
reproducible.


nfpm.go
```diff
@@ -230,3 +233,43 @@ func WithDefaults(info *Info) *Info {
 
 	return info
 }
+
+// CopiedFile describes the source and destination of one file to copy into a
+// package and whether it is a config file
```
caarlos0, line 238:
```suggestion
// package and whether it is a config file.
```

nfpm.go
```diff
@@ -230,3 +233,43 @@ func WithDefaults(info *Info) *Info {
 
 	return info
 }
+
+// CopiedFile describes the source and destination of one file to copy into a
```
caarlos0, line 237: the name is a bit misleading... its not a copied file, is a file to be copied... 🤔

dzhu: Good point. How about "FileToCopy" (and "FilesToCopy" for the method)?

caarlos0:

Issue #133, opened 2020-05-10
Support for Arch Linux packages?


dzhu:
Would there be any interest in adding support for [Arch Linux](https://www.archlinux.org) packages? I _may_ be able to provide this myself (no promises), so, even though I'm guessing it won't be a high priority, it would be good to know whether the answer is "yes, if there is a PR" or "no, we wouldn't want that".

caarlos0:
check https://github.com/goreleaser/nfpm/pull/126

closing as duplicated

dzhu:
Er, I don't think this is a duplicate: this is about Arch Linux, rather than Alpine Linux. Looks like there's a bit of [overlap](https://wiki.alpinelinux.org/wiki/Alpine_package_format#.PKGINFO) in the format of `.PKGINFO`, which I didn't know before, but otherwise they involve different formats.

caarlos0:
sorry, totally misread :|

dzhu:
No worries. :slightly_smiling_face: I'll look into doing this, then.

olanod:
This would be awesome, any updates with the PR?

caarlos0:
fwiw, goreleaser can now generate AUR PKGBUILD files and push to a remote, no packages directly, though.

https://goreleaser.com/customization/aur/

PR #141, opened 2020-05-16
feat: add support for Arch Linux packages

dzhu:
## Description

This adds support for creating [Arch Linux](https://www.archlinux.org) packages.

A minimal Arch Linux package is quite simple: it's a tarball containing the installed contents of the package plus one extra file for metadata.

The only required metadata file is `.PKGINFO`, which contains core metadata about the package: name, version, dependencies, etc.

Install hooks, if present, are stored in `.INSTALL`, which is a shell script containing a function for each phase (pre-install, etc.); this is unlike deb and RPM, which have one file per phase.

There is also usually `.MTREE`, which contains a summary of the files in the package in mtree format. It is not required to install a package, but it's needed for determining whether any files have been changed after installation, which is pretty important.

The above files are created by this code.

Another common file is `.BUILDINFO`, containing some extra metadata about the environment in which the package was built; that is not created.

Also, packages are typically compressed with zstd (the current standard) or xz (the previous one); there doesn't seem to be a high-quality, pure-Go library for either of those, however, so we stick to the Go standard library and use gzip, which still works fine.

This commit also introduces some helper code that should be useful for other package formats, but we leave any potential refactors for the future.

## Notes

Tests are minimal at the moment; I imagine there will need to be more before this can be merged, but I wanted to get through any updates to the main code before getting too far into testing. I've been doing end-to-end testing manually, and everything is already functioning, as far as I can tell.

One wrinkle I'm not sure how to handle is that there is no dedicated file extension for Arch packages, unlike with `.deb` or `.rpm`; they're simply named `pkgname.pkg.tar` plus the extension for the compression algorithm—e.g., `pkgname.pkg.tar.zstd` or, in this case, `pkgname.pkg.tar.gz`. So currently I have the packager registered under the name `gz`, but that seems like a silly state of affairs.

Closes: #133


arch/arch.go
```diff
@@ -0,0 +1,209 @@
...
+	})
+	if err := template.Must(tmpl.Parse(pkgInfoTemplate)).Execute(buf, controlData{info, instSize}); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+func addInstallScript(info *nfpm.Info, tarFile *tarhelper.File) error {
+	if info.ArchLinux.InstallScript == "" {
```
caarlos0, line 179: we'll need to add to the docs that the `scripts` field is not effective on arch packages... 

or maybe we can use one field from there? not sure what's less confusing 🤔

nfpm.go
```diff
@@ -152,6 +152,7 @@ type Overridables struct {
 	Scripts      Scripts           `yaml:"scripts,omitempty"`
 	RPM          RPM               `yaml:"rpm,omitempty"`
 	Deb          Deb               `yaml:"deb,omitempty"`
+	ArchLinux    Arch              `yaml:"gz,omitempty"`
```
caarlos0, line 155: Maybe use `ArchPKG` or just `pkg` everywhere?

caarlos0:
Hey @dzhu, thanks so much for the PR!

Overall itt looks good to me :) I made a couple of comments bellow and on the code itself, lmk what you think :)

For this to be merged in, can you update the docs and add the acceptance tests?

Thanks a ton 🚀

---

about your inquiries:

> One wrinkle I'm not sure how to handle is that there is no dedicated file extension for Arch packages, unlike with .deb or .rpm; they're simply named pkgname.pkg.tar plus the extension for the compression algorithm—e.g., pkgname.pkg.tar.zstd or, in this case, pkgname.pkg.tar.gz. So currently I have the packager registered under the name gz, but that seems like a silly state of affairs.

I agree, the extension thing is not ideal. I'll make the needed changes to the binary so we can register as the format, which in this case I think could be `pkg` or `archpkg` or something like that. Will let you know :)

> This commit also introduces some helper code that should be useful for other package formats, but we leave any potential refactors for the future.

that is nice and we can for sure refactor at some point... for now maybe leave it inside the arch pkg? We can always extract it later as we see fit :)

nxadm:
I see that [Arch supports GPG signed packages](https://wiki.archlinux.org/index.php/Pacman/Package_signing), like the other packages supported by nfpm. Does the PR takes this into account?

djgilcrease:
> I see that [Arch supports GPG signed packages](https://wiki.archlinux.org/index.php/Pacman/Package_signing), like the other packages supported by nfpm. Does the PR takes this into account?

This PR is fairly out of date and would need to be updated to the new format for packagers

caarlos0:
closing due to lack of activity.

PR #152, opened 2020-07-01
fix rpm recommends

rbtr:
Fixes goreleaser/goreleaser/issues/1636

`recommends` was not being included in the rpmpack struct so including that field in the nfpm config was ineffective. fixed that so that rpms correctly include the recommended packages.

After applying this patch and rebuilding goreleaser:
```
$ sudo dnf repoquery --recommends ./my-app.rpm
Last metadata expiration check: 2:10:04 ago on Wed 01 Jul 2020 10:28:08 AM EDT.
< the recommended dependency >

$ sudo dnf install ./my-app.rpm               
Last metadata expiration check: 2:10:11 ago on Wed 01 Jul 2020 10:28:08 AM EDT.
Dependencies resolved.
===
[...]
Installing weak dependencies:
 < the recommended dependency >
```

### DIFF
--- a/acceptance/testdata/complex.yaml
+++ b/acceptance/testdata/complex.yaml
@@ -7,6 +7,8 @@ depends:
 - bash
 provides:
 - fake
+recommends:
+- fish
 replaces:
 - foo
 suggests:
--- a/acceptance/testdata/rpm.complex.dockerfile
+++ b/acceptance/testdata/rpm.complex.dockerfile
@@ -1,6 +1,8 @@
 FROM fedora
 ARG package
 COPY ${package} /tmp/foo.rpm
+RUN test "$(rpm -qp --recommends /tmp/foo.rpm)" = "fish"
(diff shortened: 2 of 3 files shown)


caarlos0:
awesome, thanks for the PR!

do you mind adding tests as well? 

I think we can add recommends [here](https://github.com/goreleaser/nfpm/blob/master/acceptance/testdata/complex.yaml) and an assertion [here](https://github.com/goreleaser/nfpm/blob/master/acceptance/testdata/rpm.complex.dockerfile)...

rbtr:
Updated. Since the test package is installed with `rpm`, dependencies are not automatically resolved - it would need to use yum or dnf to automatically install those. The best we can do instead is check that the rpm self-reports its `recommends` correctly. I added that and a similar test for the existing `suggests`.
I unfortunately can't figure out an equivalent way to do this in dpkg, so I have left the .deb test alone.

caarlos0:
awesome, thanks!

caarlos0:

PR #157, opened 2020-07-09
WIP: feat: Use conventional file names by default

erikgeiser:
This pull request changes the behaviour of the `--target` CLI option to produce conventional file for the respective packages by default. Conventional file names are for example [package_version_architecture.package-type](https://manpages.debian.org/buster/dpkg-dev/dpkg-name.1.en.html) for Debian packages and [name-version-release.architecture.rpm](http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html) for RPMs. Consider the following cases:
* **`--target` is blank or omitted:** Create a package with a conventional file name in current directory
* **`--target` is a directory:** Create a package with a conventional file name in the directory
* **`--target` is a file name:** Use this file name

### Why?
Because at the moment, `nfpm` allows users to omit `--target` and uses the very weird default `/tmp/foo.deb` in this case. A default value should cover the most common use cases and creating the file `/tmp/foo.deb` would be a very weird use case and is just plain wrong for RPM files.

### At the moment, there are a few issues:
* I don't know if adding a new function to the Packager interface is a bit too drastic, but currently, I cannot move the function to `nfpm.go` (where I initially put it) because it needs access to the architecture translation and it has to be ensured that every future architecture provides such a name.
* I also changed a bit at the end of `doPackage()` because the success message now has to be printed there. At the same time, I changed the behaviour to remove the empty package file if packaging fails. Is it okay to also include that in this pull request? 
* I could not run the tests because I don't have access to docker on this machine

### DIFF
--- a/cmd/nfpm/main.go
+++ b/cmd/nfpm/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 
 	"github.com/alecthomas/kingpin"
@@ -25,8 +26,8 @@ var (
 		String()
 
 	pkgCmd = app.Command("pkg", "package based on the config file").Alias("package")
-	target = pkgCmd.Flag("target", "where to save the generated package").
-		Default("/tmp/foo.deb").
+	target = pkgCmd.Flag("target", "where to save the generated package (filename, folder or blank for current folder)").
+		Default("").
 		Short('t').
 		String()
 	packager = pkgCmd.Flag("packager", "which packager implementation to use").
@@ -50,7 +51,6 @@ func main() {
 		if err := doPackage(*config, *target, *packager); err != nil {
 			kingpin.Fatalf(err.Error())
 		}
-		fmt.Printf("created package: %s\n", *target)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


caarlos0:
> * I don't know if adding a new function to the Packager interface is a bit too drastic, but currently, I cannot move the function to `nfpm.go` (where I initially put it) because it needs access to the architecture translation and it has to be ensured that every future architecture provides such a name.

IMHO its OK.

> * I also changed a bit at the end of `doPackage()` because the success message now has to be printed there. At the same time, I changed the behaviour to remove the empty package file if packaging fails. Is it okay to also include that in this pull request?

I think we can do it in another one, since this one is pretty much done...

> * I could not run the tests because I don't have access to docker on this machine

no problem, CI passed :) although there are some missing coverage, but we can handle that later..

erikgeiser:
> I think we can do it in another one, since this one is pretty much done...

This is already included in this pull request because I had to change this part due to the call to `fmt.Printf` after `pkg.Package()`. So now we have these options:

**A: Before this pull request**
```go
	return pkg.Package(info, f)
```
**B: Currently in this pull request**
```go
	err = pkg.Package(info, f)
	_ = f.Close()
	if err != nil {
		os.Remove(target)
		return err
	}

	fmt.Printf("created package: %s\n", target)
	return nil
```
**C: If I remove the file removal again:**
```go
	err = pkg.Package(info, f)
	if err != nil {
		return err
	}

	fmt.Printf("created package: %s\n", target)
	return nil
```

What do you prefer for this pull request?

caarlos0:
Let's leave as is then.

Can I merge it?

erikgeiser:
> Can I merge it?

As far as I'm concerned, sure.

caarlos0:
awesome, thanks for the PR :)

caarlos0:

PR #158, opened 2020-07-09
feat(#42): Support Changelogs

erikgeiser:
This pull request adds the ability to add a changelog by specifying a [](https://github.com/goreleaser/chglog) compatible YAML file in the configuration file. This pull request was built based on the [djgilcrease's great pull request #105](https://github.com/goreleaser/nfpm/pull/105), however, the following changes were made:
* All properties of the changelog are entirely defined by the changelog YAML file. No default changelog entry is created and no information is sourced from `Info`. `nfpm` should not manage the changelog. Instead, it should only provide the means to add your own changelog to a package.
* If no changelog YAML file is specified, nothing changelog-related is added to the package for users that explicitly do not want a changelog.
* Dependency from specifc pull request branch of the `rpmpack` repos was removed. This was necessary because the pull request was closed. A different API for the same purpose was added since, so I used this API.
* The nomenclature for the configuration file was change from `chglog` to `changelog` to be more clear for a better user experience.
* The pull request is compatible with `master` again.
* I did not add any changelog information for the `nfpm` package itself, because this pull request only concerns the feature support and maintenance of the `nfpm` package itself is kind of orthogonal to that.


erikgeiser:
Hmm, `go generate -v .` produces the following error:
```
# github.com/golangci/golangci-lint/pkg/golinters
../go/pkg/mod/github.com/golangci/golangci-lint@v1.28.1/pkg/golinters/gofumpt.go:49:41: cannot use "" (type string) as type format.Options in argument to format.Source
nfpm.go:1: running "go": exit status 2
```
As far as I can see, compared to `master`, this regression was introduced by adding the dependency https://github.com/goreleaser/chglog. I have no idea why. I could also easily get rid of the dependency. All I really need are the structures to unmarshal YAML and the deb template. What do you think?

erikgeiser:
I added in depth tests for the `./deb` and `./rpm` packages as well as acceptance tests for both formats. I also updated the documentation. From my side, this is now ready to be merged.

caarlos0:
I'll review it later today or over the weekend, thanks for the PR meanwhile :)

erikgeiser:
I just notices that I most of my commits with a wrong mail address. Would it be okay to squash all these messy commits into a single correct one and force push?

caarlos0:
> I just notices that I accidentally sent most of my commits with the wrong mail address. Would it be okay to squash all these messy commits into a single correct one and force push?

yes, no problem :)

caarlos0:
Also would be nice to include @djgilcrease as co-author

djgilcrease:
Thank you very much, I got pulled to a new project at work that has kept me too busy to get back to this. This looks great.

erikgeiser:
I cleaned up my commits.

> Also would be nice to include @djgilcrease as co-author

How can I do this, or is this something that's done when merging into master? I already tried to give credit in the commit message and in the description above, but being a co-author is obviously nicer.

djgilcrease:
> How can I do this, or is this something that's done when merging into master? I already tried to give credit in the commit message and in the description above, but being a co-author is obviously nicer.

He just wanted me to take a look and ensure the direction is still where I wanted to go with it. I like the changes you made to have it so nfpm only uses the file if it is there but leaves the generation and management up to the chglog tool. If we want tighter integration in the future we can add config option to enable auto generation of changelog.

erikgeiser:
> I like the changes you made to have it so nfpm only uses the file if it is there but leaves the generation and management up to the chglog tool. If we want tighter integration in the future we can add config option to enable auto generation of changelog.

I did this because I envision `nfpm` to be tailor-made tool precisely to craft package files. In my opinion, changelogs are mostly orthogonal to that idea other than the fact that the package formats require special integration of changelog files. That's why I think `nfpm` should not manage changelogs but only provides means to include them in package formats correctly.

But maybe my point of view is a bit biased because I primarily use `nfpm` wrapped by out own custom build system tailored to out needs such that `nfpm` is simple the "assembly tool" for package files.

djgilcrease:
> But maybe my point of view is a bit biased because I primarily use `nfpm` wrapped by out own custom build system tailored to out needs such that `nfpm` is simple the "assembly tool" for package files. This strategy would also help to keep it from getting as bloated as `fpm`.

This is how I actually use it as well, just a stage in the pipeline to package things to be pushed up to an internal artifactory server.

caarlos0:
> How can I do this, or is this something that's done when merging into master?

when you rebase the commits, just add 

```
Co-authored-by: another-name <another-name@example.com>"
```

to the end of the commit message https://docs.github.com/en/enterprise/2.13/user/articles/creating-a-commit-with-multiple-authors

erikgeiser:
> when you rebase the commits, just add to the end of the commit message :)

Just for my curiousity: I did `git rebase master; git commit --amend` and then had to force push again. Is there a better way to do this?

caarlos0:
> Just for my curiousity: I did `git rebase master; git commit --amend` and then had to force push again. Is there a better way to do this?

AFAIK that's it...

caarlos0:

Issue #125, opened 2020-03-25
feature request: allow empty `files:`


nickatsegment:
This might be out of scope for nfpm, but I want to use it to make a metapackage with no files. Currently it requires the `files` field to have some contents.

erikgeiser:
I don't think that's out of scope at all. I think it would be a great feature.

caarlos0:
This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue for related bugs.

PR #159, opened 2020-07-09
feat: allow creating meta packages

caarlos0:
closes https://github.com/goreleaser/nfpm/issues/125

---

works on debs, but fails on rpms:

```
error: foo.rpm: tag[17]: BAD, tag 1028 type 4 offset 268 count 0 len 0
error: foo.rpm cannot be installed
```

need to investigate why


nickatsegment:
Thanks @caarlos0!

caarlos0:

PR #178, opened 2020-07-25
feat: Add support for DEB triggers.

erikgeiser:
This pull request adds support for triggers in .deb files. These triggers are commonly used for tasks like rebuilding man pages or `initramfs`. Packages can activate a trigger for other packages or claim interest on triggers activated by other packages. For further information see https://wiki.debian.org/DpkgTriggers, https://man7.org/linux/man-pages/man5/deb-triggers.5.html.

In addition to the feature support, this pull request adds **unit tests**, an **acceptance test** and **documentation**. However, the following changes probably need some feedback before merging:
* The trigger names internally use a `-` (e.g. `activate-noawait`), however I used a `_` (e.g. `activate_noawait`)for the `nfpm` configuration to adhere to yaml conventions.
* `createControl` is now 2 or 3 lines above the line limit enforced by `golangci-lint:funlen`. I think it is still perfectly readable and simple enough, so I decided to ignore the warning by adding `// nolint:funlen`. I hope that's okay.
* In the acceptance test, the package only claims interest for one trigger which is then triggered manually. Activating a trigger is not explicitly tested. The main reason is that `accept()` currently does not support building two packages for the Docker container which would be necessary to test both ways of the trigger mechanism. However, the mechanism is simple enough that I think this test is sufficient, especially since the other cases are somewhat covered by the unit test.

### DIFF
--- a/www/docs/configuration.md
+++ b/www/docs/configuration.md
@@ -129,11 +129,11 @@ deb:
   # Custom deb triggers
   triggers:
     # register interrest on a trigger activated by another package
-    # (also available: interest-await, interest-noawait)
+    # (also available: interest_await, interest_noawait)
     interest:
       - some-trigger-name
     # activate a trigger for another package
-    # (also available: activate-await, activate-noawait)
+    # (also available: activate_await, activate_noawait)
     activate:
       - another-trigger-name
 ```
--- a/www/docs/configuration.md
+++ b/www/docs/configuration.md
@@ -125,6 +125,17 @@ deb:
   # Custom deb rules script.
   scripts:
     rules: foo.sh
+
+  # Custom deb triggers
+  triggers:
+    # register interrest on a trigger activated by another package
+    # (also available: interest-await, interest-noawait)
+    interest:
+      - some-trigger-name
+    # activate a trigger for another package
+    # (also available: activate-await, activate-noawait)
+    activate:
+      - another-trigger-name
 ```
 
 ## Templating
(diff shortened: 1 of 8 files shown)


www/docs/configuration.md
```diff
@@ -125,6 +125,17 @@ deb:
...
     rules: foo.sh
+
+  # Custom deb triggers
+  triggers:
+    # register interrest on a trigger activated by another package
+    # (also available: interest-await, interest-noawait)
+    interest:
+      - some-trigger-name
+    # activate a trigger for another package
+    # (also available: activate-await, activate-noawait)
```
caarlos0, line 136: if I understand correctly:

```suggestion
    # (also available: activate_await, activate_noawait)
```

www/docs/configuration.md
```diff
@@ -125,6 +125,17 @@ deb:
   # Custom deb rules script.
   scripts:
     rules: foo.sh
+
+  # Custom deb triggers
+  triggers:
+    # register interrest on a trigger activated by another package
+    # (also available: interest-await, interest-noawait)
```
caarlos0, line 132: I think it should be:

```suggestion
    # (also available: interest_await, interest_noawait)
```

erikgeiser:
Yes, fixed it. I did a last minute change to adhere to the yaml conventions and I missed the docs.

caarlos0:
can you resolve the conflicts as well? other than that I think we are good to merge :)

erikgeiser:
Done.

caarlos0:
thanks!

caarlos0:

Issue #132, opened 2020-05-05
On macOS nfpm cannot read symlinks


nhooyr:
```
name: "code-server"
arch: "${ARCH}"
platform: "linux"
version: "v${VERSION}"
section: "devel"
priority: "optional"
maintainer: "Anmol Sethi <hi@nhooyr.io>"
description: |
  Run VS Code in the web browser.
vendor: "Coder"
homepage: "https://github.com/cdr/code-server"
license: "MIT"
bindir: "/usr/bin"
files:
  ./nfpm-code-server-symlink: /usr/bin/code-server
  ./release-static/**/*: "/usr/lib/code-server/"
```

The first file in the config is a symlink to `/usr/lib/code-server/code-server` which is where the main binary for my package is.

The deb builds fine on linux but on Mac, it fails with

```
using deb packager...
nfpm: error: ./nfpm-code-server-symlink: file does not exist
```

nhooyr:
On linux actually the symlink is followed and the real file is copied directly.

erikgeiser:
I could not reproduce this behaviour on my MacBook running Catalina. However, I created a pull requests with tests as starting point for further investigation.

caarlos0:
closed by mistake

erikgeiser:
@nhooyr Symlink handling changed quite a bit in `nfpm` since you opened this issue.

* **Symlinks in `files`:** Symlinks in files followed and the actual file is copied to the destination. As far as I am aware and according to my own tests, this also works the same way on macOS (I traced the file read operation up to the final syscall on macOS which should work the same as on Linux). As a result, `files` should not be used if you want an actual symlink in the package.
* **Symlinks in `symlinks`:** In `symlinks` you can define `src: dst` mappings which will be create symlinks inside the package. Both `src` and `dst` are paths in the package's root filesystem. Thus there is no need to "read" a symlink from disk on the build host and as a result, this method is build operating system agnostic.

Do you still encounter an issue with macOS? If not, can this issue be closed?

nhooyr:
Yea this was likely fixed. If I encounter it again, I'll reopen!

caarlos0:
Nice, thanks everyone!

caarlos0:
This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue for related bugs.

PR #180, opened 2020-07-25
test(#132): Add tests with symlinked files.

erikgeiser:
I tried to fix #132, where the build failed for configurations in which files are referenced via symlinks in macOS. So first I wrote some tests for both `rpm` and `deb` packages to reproduce the issue. However, on my MacBook running Catalina, the packages built perfectly fine and symlinks were followed, so I did not change anything in the packaging process.

I figured, more tests are never bad so this pull request contains the tests with symlinks. If I missed something and the issue does indeed exist, these tests are also still a great starting point for further debugging. Also the tests introduce a handy helper function to extract files from `rpm`s and they also fix a minor issue with the helper function to extract files from `.tar.gz` archives.

PR #181, opened 2020-07-27
fix: Do not expand symlink on deb.

rsur:
This prevents symlinks from being expanded. For deb only.

See #132

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -202,36 +202,58 @@ func createEmptyFoldersInsideTarGz(info *nfpm.Info, out *tar.Writer, created map
 }
 
 func copyToTarAndDigest(tarw *tar.Writer, md5w io.Writer, src, dst string) (int64, error) {
-	file, err := os.OpenFile(src, os.O_RDONLY, 0600) //nolint:gosec
+	stt, err := os.Lstat(src)
 	if err != nil {
 		return 0, errors.Wrap(err, "could not add file to the archive")
 	}
+	var linkname *string
+	switch mode := stt.Mode(); {
+	case mode&os.ModeSymlink != 0:
+		lpath, _ := filepath.EvalSymlinks(src)
+		if lpath[0] != '/' {
+			// relative path is re-resolved relative to src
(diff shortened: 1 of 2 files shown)


erikgeiser:
I cannot reproduce #132 on macOS Calatina and I'm not sure the changes here are effective. As far as I know `os.OpenFile` follows symlinks unless this behaviour is explicitly disabled, so I don't really see the need to manually inspect the file. My pull request #180 was just merged which adds tests for symlinked files. We should find a setup in which these tests fail first.

caarlos0:
> I cannot reproduce #132 on macOS Calatina and I'm not sure the changes here are effective. As far as I know `os.OpenFile` follows symlinks unless this behaviour is explicitly disabled, so I don't really see the need to manually inspect the file. My pull request #180 was just merged which adds tests for symlinked files. We should find a setup in which these tests fail first.

notice the tests are running on linux (on CI) and are also passing there...

erikgeiser:
Sure, but I also ran them on my MacBook.

caarlos0:
> Sure, but I also ran them on my MacBook.

yes, I meant that the tests are passing on both linux and mac :)

erikgeiser:
I did a bit of research and this is the callstack:
 [`os.OpenFile` calls `openFileNolog`](https://golang.org/src/os/file.go?s=9377:9443#L295)
[`openFileNolog` on `unix` and `darwin` calls `syscall.Open`](https://golang.org/src/os/file_unix.go#L189)
[`syscall.Open` on `dawin` is directly exposed](https://golang.org/src/syscall/syscall_darwin.go#L299)

and `open` on macOS, according to the [man pages](http://www.manpagez.com/man/2/open/osx-10.6.php) follows symlink except for two special cases:
* `O_NOFOLLOW` is set: In this case symlinks are explicitly **not** followed
* `O_SYMLINK` is set: In this case the symlink **itself** is opened

Both flags are neither set by the Go standard lib, nor by `nfpm`. Therefore, I see no reason why symlinks shouldn't work on macOS.

rsur:
Resumed to #184.

caarlos0:

PR #184, opened 2020-07-27
fix: Do not expand symlink on deb.

rsur:
See #183 

For deb only.

PR #185, opened 2020-07-28
feat(#183): Support symlinks for deb and rpm.

### LINKED ISSUE #183: Do not expand symlink on packaged file.
Per deb [specification](https://www.debian.org/doc/debian-policy/ch-files.html#symbolic-links), not only unexpanded (i.e. left as is) symlink is allowed to be shipped, it can even point to non-existent file on the source package tree, assuming the target file exists after installation. Relative symlink is strongly suggested, but absolute symlink is not entirely forbidded.

This behavior is observable on `fpm`.

So, the assertion [here](https://github.com/goreleaser/nfpm/pull/180/files#diff-363799588117b0eb7767f12380fd7afeR526) is inaccurate even though a temporary symlink is created in place of `../testdata/whatever.conf`, because it compares the content of source and packed files. It always assumes that the packed file is regular, not a symlink.

For the sake of completion, we can do tests based on this toy YAML:

```yaml
name: "some-package"
arch: "amd64"
version: "1.0.0"
files:
  ./bin/big-program: "/usr/bin/big-program"
  ./bin/sym-program: "/usr/bin/sym-program"
```

where `./bin/sym-program` is a relative symlink to `./bin/big-program`, i.e. `big-program`, hence `/usr/bin/sym-program` must also be a relative symlink to `/usr/bin/big-program`, i.e. `big-program`. We only need to ensure `/usr/bin/sym-program` has linkname `big-program` and resolve it to `/usr/bin/big-program`. We can compare the content from there on.

erikgeiser:
This pull request is an alternative approach to #184 for adding symlinks, as described in my comments in issue #183 (closes #183). In this approach, symlinks can be added as follows:
```
symlinks:
  /path/to/symlink: /path/to/target
```

This approach will not break existing build configurations and supports both package formats. The API is consistent with `config_files` and `empty_folders` and does not promote symlinks in `files` which are broken on the build host that will only eventually work in the context of the package root. Also the changes to the `nfpm.go`, `deb.go` and `rpm.go` are very minimal. Also the separate configuration option makes it clear that symlinks `files:` are followed because otherwise there would be no need for a separate option.

 If we want to develop this approach, I will add tests and documentation. (EDIT: already done)

To discuss:
* In the approach in #184, the symlink target file size is added to the `instSize` for `deb` packages. In this approach nothing is added to `instSize` as symlinks do not have a size (`du -sh` reports `0`). I don't know which is the better option.

### DIFF
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -214,7 +214,7 @@ func TestSymlink(t *testing.T) {
 		t.Run(fmt.Sprintf("symlink-%s", format), func(t *testing.T) {
 			t.Parallel()
 			accept(t, acceptParms{
-				Name:       fmt.Sprintf("changelog_%s", format),
+				Name:       fmt.Sprintf("symlink_%s", format),
 				Conf:       "symlink.yaml",
 				Format:     format,
 				Dockerfile: fmt.Sprintf("%s.symlink.dockerfile", format),
--- a/acceptance/acceptance_test.go
+++ b/acceptance/acceptance_test.go
@@ -208,6 +208,21 @@ func TestDebTriggers(t *testing.T) {
 	})
 }
 
+func TestSymlink(t *testing.T) {
+	for _, format := range formats {
+		format := format
+		t.Run(fmt.Sprintf("symlink-%s", format), func(t *testing.T) {
+			t.Parallel()
+			accept(t, acceptParms{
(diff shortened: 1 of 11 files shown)


erikgeiser:
I got bored already added tests and docs ;)

caarlos0:
Thanks!

caarlos0:

PR #186, opened 2020-07-29
feat(#129): Support RPM %config(noreplace) files.

### LINKED ISSUE #129: cant support config(noreplace)
for upgrade a rpm,  the config(noreplace) is an important feature.

erikgeiser:
This pull request closes #129 by introducing the RPM-specific configuration option `config_noreplace_files` that works like `config_files` but creates files that will not be overwritten by new versions if they were changed by the system administrator. This corresponds to the [RPM spec option `%config(noreplace)`](https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html).

Unit test, acceptance tests and documentation is included. However, I had to make changes to the exported `nfpm.FilesToCopy` function and to the `nfpm.FileToCopy` struct to make them more flexible for changes related to new types of files, this is probably bad if someone relies on the old API. Also the acceptance test is a bit dirty because I needed two different packages inside the docker container. I hope that's okay.


erikgeiser:
For this pull request, I also created an issue for `rpmpack` (https://github.com/google/rpmpack/issues/47). It this is implemented, this pull request could be cleaned up a bit.

Otherwise, it comes down to question whether or not the API changes of exported objects is acceptable for `nfpm`. I don't know if `nfpm` is actively being used as a library. If not, I think the changes are beneficial because they allow for more flexibility for future changes with different file handling between package formats.

caarlos0:
> For this pull request, I also created an issue for `rpmpack` ([google/rpmpack#47](https://github.com/google/rpmpack/issues/47)). It this is implemented, this pull request could be cleaned up a bit. If this PR is merged sooner, the cleanup could of course also be done in another PR later on.
> 
> Otherwise, it comes down to question whether or not the API changes of exported objects is acceptable for `nfpm`. I don't know if `nfpm` is actively being used as a library. If not, I think the changes are beneficial because they allow for more flexibility for future changes with different file handling between package formats.

I'll open a PR there...

erikgeiser:
Cool ;). I also just realized that my second paragraph was a bit ambiguous. In the second paragraph I meant the API changes in `nfpm.go` introduced in this pull request.

caarlos0:
https://github.com/google/rpmpack/pull/48

erikgeiser:
I did a rebase and incorporated the upstream changes from google/rpmpack#48.

acceptance/testdata/config-noreplace-old.yaml
```diff
@@ -0,0 +1,13 @@
+name: "foo"
+arch: "all"
+platform: "linux"
+version: "v1.0.0"
+maintainer: "Foo Bar"
+vendor: "foobar"
+homepage: "https://foobar.org"
+license: "MIT"
+config_files:
+  ../testdata/whatever.conf: "/etc/regular.conf"
```
caarlos0, line 10: I was thinking, maybe we could change this thing to be a `map[string]struct`, and then make it something like:

```yaml
config_files:
   ../testdata/whatever.conf:
     path: "/etc/regular.conf"
     replace: true
```

wdyt?

caarlos0: although, not sure this can be changed in debs 🤔

erikgeiser: According to https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html (5.3, paragraph 4) debs handle this by interrupting the upgrade to query the user about what should happen with the configuration file.

While your approach is quite elegant if both package types had the same concept, I think it is not really suitable for the following reasons:
* It breaks existing configurations
* Future package types could handle it differently, which is confusing for the user when config options may or may not have an effect depending on the package type
* We should not make the configuration more complicated to accommodate for an edge case. I think most people are fine with the regular `config_files` so only those who have a particular use case and know what they are doing can opt for different behaviour by using the separate `config_noreplace_files` option

nfpm.go
```diff
@@ -245,34 +247,31 @@ func WithDefaults(info *Info) *Info {
...
+// copy into a package.
 type FileToCopy struct {
 	Source      string
 	Destination string
-	Config      bool
 }
 
 // FilesToCopy lists all of the real files to be copied into the package.
-func (info *Info) FilesToCopy() ([]FileToCopy, error) {
+func (info *Info) FilesToCopy(srcDstGlobs map[string]string) ([]FileToCopy, error) {
```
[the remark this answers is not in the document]
caarlos0: yeah, and major bumps on go mods are really painful...

maybe we can keep this function with the old behavior and create a new one with the new signature?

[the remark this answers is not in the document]
caarlos0: actually, this function should probably be in an internal package, as its not supposed to be used directly by nfpm users... will refactor for it and open a PR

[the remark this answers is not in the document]
caarlos0: check https://github.com/goreleaser/nfpm/pull/191

internal/files/files.go
```diff
@@ -1,45 +1,32 @@
...
-			globbed, err := glob.Glob(srcglob, dstroot)
-			if err != nil {
-				return nil, err
-			}
-			for src, dst := range globbed {
-				// avoid including a partial file with the name of the target in the target
-				// itself. when used as a lib, target may not be set. in that case, src will
-				// always have the empty sufix, and all files will be ignored.
-				// TODO: add tests cases for this
-				if info.Target != "" && strings.HasSuffix(src, info.Target) {
```
erikgeiser, line 34: As a changed the signature I had to think about if I still want to pass in the target to do this check. But then I thought we should remove this check because, in my opinion, it has no business being here. If I want to include a file that ends with "deb" or "rpm" in my package I should be able to. If I accidentally glob files I do not want, that's a user error which should not be corrected magically by some heuristic in `nfpm`.

Some examples: I could have a file named `harddrives_rpm` containing rounds per minute specs of my hard drives. Or apparently `ardeb` is a unit of dry measure used in egypt (wow, this is getting a bit far fetched, sorry 😄). Both files could never end up in an package.

internal/files/files.go
```diff
@@ -1,45 +1,32 @@
...
 type FileToCopy struct {
 	Source      string
 	Destination string
-	Config      bool
 }
 
-// ToCopy lists all of the real files to be copied into the package.
-func ToCopy(info *nfpm.Info) ([]FileToCopy, error) {
+// Expand gathers all of the real files to be copied into the package.
+func Expand(filesSrcDstMap map[string]string) ([]FileToCopy, error) {
```
erikgeiser, line 17: I think Expand is a fitting name since this function basically just expands the globs in a source->dest map.

internal/files/files.go
caarlos0: makes sense to me :)

internal/files/files.go
caarlos0: agreed!

caarlos0:
thanks for the work on this (and all other PRs) @erikgeiser ! 🙏🚀

erikgeiser:
Thanks for the fast feedback always @caarlos0, it's a pleasure to help out

caarlos0:

PR #187, opened 2020-07-29
fix(#130): Don't propagate semver metadata into deb metadata.

### LINKED ISSUE #130: build metadata in .deb version breaks precedence order
### Issue
According to [deb-version](http://man7.org/linux/man-pages/man5/deb-version.5.html), all version sections are taken in consideration while calculating the version precedence order. However, `nfpm` adds the semver build metadata to the .deb package version, e.g. semver `0.1.11-beta3+deadbeef` becomes .deb version `0.1.11~beta3+deadbeef`. This causes dpkg to sort package versions in weird order.

This could be confirmed using `dpkg --compare-versions` (tested on Ubuntu 18.04.3):
```
% dpkg --compare-versions 0.1.11~beta3+12345678 eq 0.1.11~beta3+deadbeef && echo '=' || echo '!='
!=
% dpkg --compare-versions 0.1.11~beta3+12345678 lt 0.1.11~beta3+deadbeef && echo '<' || echo '>='
<
% dpkg --compare-versions 0.1.11~beta3+deadbeef gt 0.1.11~beta3+12345678 && echo '>' || echo '<='
>
```

This behavior contradicts the semver semantics which [clearly states](https://semver.org/#spec-item-10) that
>Build metadata MUST be ignored when determining version precedence.

### Possible solution
Do not put semver build metadata into the .deb version metadata. Removing this line will do the trick. https://github.com/goreleaser/nfpm/blob/e0254b2770db441b1dbbdad0b292b65671c1f119/nfpm.go#L208

erikgeiser:
This pull request closes #130 by not propagating the semver metadata info the `Deb.Metadata` field. As described in the corresponding issue, this is the expected behaviour for `deb` files.

EDIT: changed it to wip because I actually don't even know why `Deb.Metadata` exists if it shouldn't be included in the version.

### DIFF
--- a/acceptance/testdata/deb.env-var-version.dockerfile
+++ b/acceptance/testdata/deb.env-var-version.dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu
 ARG package
 COPY ${package} /tmp/foo.deb
-ENV EXPECTVER=" Version: 1.0.0~0.1.b1+git.abcdefgh"
+ENV EXPECTVER=" Version: 1.0.0~0.1.b1"
 RUN dpkg --info /tmp/foo.deb | grep "Version" > found
 RUN export FOUND_VER="$(cat found)" && \
     echo "Expected: '${EXPECTVER}' :: Found: '${FOUND_VER}'" && \
--- a/nfpm.go
+++ b/nfpm.go
@@ -239,7 +239,6 @@ func WithDefaults(info *Info) *Info {
 		if info.Prerelease == "" && !helpers.IsInt(v.Prerelease()) {
 			info.Prerelease = v.Prerelease()
 		}
-		info.Deb.VersionMetadata = v.Metadata()
 	}
 
 	return info
--- a/nfpm_test.go
+++ b/nfpm_test.go
@@ -78,7 +78,7 @@ func TestDefaultsVersion(t *testing.T) {
 	assert.Equal(t, "1.0.0", info.Version)
 	assert.Equal(t, "2", info.Release)
 	assert.Equal(t, "beta1", info.Prerelease)
-	assert.Equal(t, "xdg2", info.Deb.VersionMetadata)
+	assert.Equal(t, "", info.Deb.VersionMetadata)
 }
 
 func TestDefaults(t *testing.T) {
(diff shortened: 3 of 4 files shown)


erikgeiser:
@caarlos0 is there still a point in keeping `Deb.Metadata` around at all if it shouldn't be included in the version?

caarlos0:
I think the user can still set something on `nfpm.yaml` 🤔

erikgeiser:
Yes but it won't end up anywhere as far as I can see. Until now in this PR I only removed the part here the metadata is populated automatically, but according to the arguments in the issue it should not end up in the packages version information at all, even if set explicitly in the config. In this case we could just remove the entire `Deb.Metadata` field altogether.

caarlos0:
> Yes but it won't end up anywhere as far as I can see. Until now in this PR I only removed the part here the metadata is populated automatically, but according to the arguments in the issue it should not end up in the packages version information at all, even if set explicitly in the config. In this case we could just remove the entire `Deb.Metadata` field altogether.

I think removing the auto-metadata seems legit, but if the user wants to do add something manually I don't see a problem...

erikgeiser:
That's a valid argument. However, the counter argument would be that this enables users to create `deb` packages where the version comparison is broken.

caarlos0:
> That's a valid argument. However, the counter argument would be that this enables users to create `deb` packages where the version comparison is broken.

It depends:

```
root@7f2aaf25ce50:/# dpkg --compare-versions 0.1.11~beta3+git eq 0.1.11~beta3+git && echo yes
yes
root@7f2aaf25ce50:/# dpkg --compare-versions 0.1.11~beta4+git gt 0.1.11~beta3+git && echo yes
yes
root@7f2aaf25ce50:/# dpkg --compare-versions 0.1.12~beta3+git gt 0.1.11~beta3+git && echo yes
yes
```

if the user wants to always set the metadata to `git`, for example, it works just fine.

The problem is that we were always changing it (setting the commit hash), and them the comparison was broken...

caarlos0:
Maybe we can just point that out in docs (to not abuse the metadata risking breaking dpkg version comparison)

erikgeiser:
Okay, let's do it this way. I changed the docs and removed the `wip`.

caarlos0:
Thanks!

caarlos0:

PR #188, opened 2020-07-31
test: Reuse rpm tag constants from rpmutils.

erikgeiser:
This pull request replace the tag constant definitions in the rpm tests with the constants that were already provided by `rpmutils`. I think that makes sense because they are always used in conjunction with `rpm.Header.GetString` which is also a `rpmutils` function. As I see it `rpmutils` is meant to be used this way.


caarlos0:
thanks

caarlos0:

PR #195, opened 2020-08-04
test: Changed some asserts to requires for early out in tests.

erikgeiser:
I changed some `assert`s to `require`s in the tests where continuing does not make sense. This should make failing tests run faster and filter the relevant error messages.


caarlos0:

PR #197, opened 2020-08-04
fix: Remove RPM group default.

### LINKED ISSUE #193: Proposed Changes to RPM Group Tag
According to https://fedoraproject.org/wiki/RPMGroups, groups are deprecated since CentOS6 and EL6 (both 9 years old). Adding a group tag is discouraged by Fedora. Additionally, I think the choice to default the group tag to `Development/Tools` is a strange default. Therefore, I propose one of the following changes:

* **Option A:** Don't set the group tag, following the current best practice. This will break the packages for CentOS5/EL5 and earlier if no group tag is set by the users. I would solve this by adding a warning to the docs that users should explicitly set the group tag if they want to support those OS versions.
* **Option B:** If we do want to keep a default group tag, we could use `Unspecified` (see https://fedoraproject.org/wiki/RPMGroups below the big list). This avoids setting a completely unrelated group for users that didn't bother setting a group (which is probably almost everyone).

If we agree how to proceed, I will submit a pull request.

erikgeiser:
This pull request removes the default RPM group tag value, adds documentation and tests for both a blank and a populated group tag. Closes #193.

Also making sure that there is always an new pull request 😆

### DIFF
--- a/rpm/rpm.go
+++ b/rpm/rpm.go
@@ -217,7 +217,7 @@ func buildRPMMeta(info *nfpm.Info) (*rpmpack.RPMMetaData, error) {
 		URL:         info.Homepage,
 		Vendor:      info.Vendor,
 		Packager:    info.Maintainer,
-		Group:       defaultTo(info.RPM.Group, "Development/Tools"),
+		Group:       info.RPM.Group,
 		Provides:    provides,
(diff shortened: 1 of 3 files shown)


caarlos0:
Thanks!

caarlos0:

PR #214, opened 2020-08-18
fix: Correctly handle version, release and prerelease for RPMs.

### LINKED ISSUE #131: Putting pre-release label into RPM Release fiels breaks precedence order
### Issue
Instead of keeping the semver pre-release label in the RPM `Version` field it's moved to the `Release` field. This goes against the RPM versioning semantics, which has support for pre-release versions since RPM v4.10 (released in 2012, see https://rpm.org/wiki/Releases/4.10.0) implemented in the very same fashion as in dpkg, using tilde (`~`) as a pre-release label delimiter.

This could be confirmed using `rpmdev-vercmp` (tested on CentOS 7.7.1908):
```
% rpmdev-vercmp 0.1.11 0.1.11~beta3
0.1.11 > 0.1.11~beta3
% rpmdev-vercmp 0.1.11~alpha1 0.1.11~beta3
0.1.11~alpha1 < 0.1.11~beta3
```
### Possible solution
Default RPM `Release` field to `1`, append RPM `Version` fiels with  `"~" + Info.Prerelease` suffix if Info.Prerelease is not empty.

erikgeiser:
This pull request fixes (hopefully) the version, release and prerelease handling for RPMs. Also, globally for all package formats, a semver prerelease present in `info.Version` will not populate `info.Release` anymore since `info.Version` should contain version information of the packaged subject and not the package itself.

Closes #131. Maybe we should also wait for feedback from @boogie-byte before merging.

### DIFF
--- a/acceptance/testdata/rpm.env-var-version.dockerfile
+++ b/acceptance/testdata/rpm.env-var-version.dockerfile
@@ -1,8 +1,8 @@
 FROM fedora
 ARG package
 COPY ${package} /tmp/foo.rpm
-ENV EXPECTVER="Version : 1.0.0" \
-    EXPECTREL="Release : 0.1.0.1.b1"
+ENV EXPECTVER="Version : 1.0.0~0.1.b1" \
+    EXPECTREL="Release : 1"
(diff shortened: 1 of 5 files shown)


caarlos0:
thanks!

caarlos0:

PR #215, opened 2020-08-19
feat: Introduce version_metadata for all packagers (deprecate deb.metadata)

erikgeiser:
I added the global config field `VersionMetadata` for all packagers. I chose `version_metadata` for the YAML field because metadata is such a generic term that most people won't associate it with the version otherwise. The `deb` packager already supported this and the `apk` formatter borrowed it from the `deb`-specific option which was not ideal.

`Deb.VersionMetadata` is now deprecated an a warning will be emitted to stderr if it is still set (same for `BinDir`).

If not explicitly set, the `VersionMetadata` can be populated from `Version` semver metadata.

Also the tests and conventional file names are updated.

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -77,7 +77,25 @@ type Apk struct{}
 
 func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
 	// TODO: verify this
-	return fmt.Sprintf("%s_%s_%s.apk", info.Name, info.Version, info.Arch)
+	arch, ok := archToAlpine[info.Arch]
+	if !ok {
+		arch = info.Arch
+	}
+
+	version := info.Version
+	if info.Release != "" {
+		version += "-" + info.Release
+	}
+
+	if info.Prerelease != "" {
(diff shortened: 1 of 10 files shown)


apk/apk.go
```diff
@@ -453,7 +471,7 @@ pkgname = {{.Info.Name}}
 pkgver = {{ if .Info.Epoch}}{{ .Info.Epoch }}:{{ end }}{{.Info.Version}}
          {{- if .Info.Release}}-{{ .Info.Release }}{{- end }}
          {{- if .Info.Prerelease}}~{{ .Info.Prerelease }}{{- end }}
-         {{- if .Info.Deb.VersionMetadata}}+{{ .Info.Deb.VersionMetadata }}{{- end }}
+         {{- if .Info.VersionMetadata}}+{{ .Info.VersionMetadata }}{{- end }}
```
caarlos0, line 474: oops, thanks for catching that!

caarlos0:
thanks!

caarlos0:

PR #217, opened 2020-08-24
fix: Ensure relative path prefix for deb packages

rtpt-alex:
This patch ensures that all items contained in tar.gz files in Debian
packages have a relative path prefix, so the names all start with "./".
While this is not strictly required for a Debian package to function,
detecting a changed config file fails with an error messages similar to
the following:

    tar: ./etc/default/conffilename: Not found in archive

It seems to fail extracting config files for checking if they have been
modified by the system administrator.

All reference documentation that I was able to find (e.g. in the Debian
Handbook[1]) shows that all paths contained in either data.tar.gz and
control.tar.gz have the prefix "./".

I've verified that with this change, config files works as expected.

[1] https://debian-handbook.info/browse/stable/packaging-system.html#sect.binary-package-structure

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -159,7 +159,7 @@ func createSymlinksInsideTarGz(info *nfpm.Info, out *tar.Writer, created map[str
 		}
 
 		err := newItemInsideTarGz(out, []byte{}, &tar.Header{
-			Name:     strings.TrimLeft(src, "/"),
+			Name:     relPathSlash(src),
 			Linkname: dst,
 			Typeflag: tar.TypeSymlink,
 			ModTime:  time.Now(),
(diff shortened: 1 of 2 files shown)


rtpt-alex:
Hey @erikgeiser, would you mind having a look? Thanks!

erikgeiser:
While investigating I found that the `md5sums` file will be utterly broken if a changelog is added. My bad. Unfortunately fixing the `md5sums` file does still not fix the issue with the missing changelog in the acceptance test.

erikgeiser:
can be closed, a fixed version is avialable in #218

caarlos0:
superseeded by #218

caarlos0:
Thanks for the PR 🚀

rtpt-alex:
Awesome, thanks @erikgeiser @caarlos0!

caarlos0:

PR #218, opened 2020-08-24
Fix: Fix deb file path prefixes and md5sums file

erikgeiser:
This pull request is based on #217 and ensures that files inside `.tar.gz` files of `deb` packages adhere to the same path prefix convention as regular upstream Debian packages. I created this PR to fix the broken acceptance test in #217 which was caused by the fact that `dpkg` in Ubuntu Docker images is configured to ignore changelog files but ONLY if we adhere to the path prefix conventions that are introduced in the PR.

During debugging I also found a nasty bug when generating the `md5sums` file if a changelog is added that results in a broken `md5sums` file. I included the fix and corresponding test in this PR because the changes were all lumped together after I was done debugging. Sorry for the mess.

### DIFF
--- a/acceptance/testdata/deb.changelog.dockerfile
+++ b/acceptance/testdata/deb.changelog.dockerfile
@@ -1,5 +1,9 @@
 FROM ubuntu
 ARG package
+# the dpkg configuration of the docker
+# image filters out changelogs by default
+# so we have to remove that rule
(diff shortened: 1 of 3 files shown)


caarlos0:
Thanks!

caarlos0:

PR #220, opened 2020-09-01
feat: Support DEB Breaks field

### LINKED ISSUE #219: Suppport "Breaks" deb relationship
It would be great if nfpm supported the [`Breaks` relationship in debian packages](https://www.debian.org/doc/debian-policy/ch-relationships.html#s-breaks)

erikgeiser:
This PR adds the `Break` field and an acceptance test. Closes #219.


erikgeiser:
Oops, I accidentally auto-formatted the config in the docs. It looks nice through, so I think we should leave it like that.

erikgeiser:
The tests are running out of time again, should I increase the limit to something like 15mins?

caarlos0:
> The tests are running out of time again, should I increase the limit to something like 15mins?

sounds good

caarlos0:

PR #223, opened 2020-09-12
fix: Change upgrade to update in changelog dockerfile

erikgeiser:
Change `apt upgrade` to `apt update` before installing a package. This was a typo and now this test should also run faster.


caarlos0:

PR #224, opened 2020-09-12
feat(#212): Package Signing

### LINKED ISSUE #212: Support Signing
add support to signing packages (on all packagers):

- deb
- rpm
- apk

erikgeiser:
* [x] `deb` signatures (`debsigs` format)
* [x] `rpm` signatures
* [x] `apk` signatures

This PR closes #212.


erikgeiser:
The build is currently failing because I did not push my `go.mod` file which contains a redirection to a local checkout of the signing PR branch of `rpmpack`.

apk/apk.go
```diff
@@ -20,7 +20,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-// Package apk (someday) implements nfpm.Packager providing .apk bindings.
+// Package apk implements nfpm.Packager providing .apk bindings.
```
caarlos0, line 23: good catch!

apk/apk.go
```diff
@@ -224,30 +239,72 @@ func createData(dataTgz io.Writer, info *nfpm.Info, sizep *int64) ([]byte, error
...
+		if err != nil {
+			return err
+		}
+
+		// needs to exist on the machine during installation: /etc/apk/keys/<keyname>.rsa.pub
+		keyname := info.APK.SigningKeyName
+		if keyname == "" {
+			addr, err := mail.ParseAddress(info.Maintainer)
+			if err != nil {
+				return errors.Wrap(err, "key name not send and unable to parse maintainer mail address")
```
caarlos0, line 272: maybe just `unable to parse maintainer email address`?

apk/apk.go
```diff
@@ -224,30 +239,72 @@ func createData(dataTgz io.Writer, info *nfpm.Info, sizep *int64) ([]byte, error
...
+			Name: fmt.Sprintf(".SIGN.RSA.%s", keyname),
+			Mode: 0600,
+			Size: int64(len(signature)),
+		}
+
+		if err := writeFile(tw, signHeader, bytes.NewReader(signature)); err != nil {
+			return err
+		}
+
+		return nil
```
caarlos0, line 288: can just:
```suggestion
		return writeFile(tw, signHeader, bytes.NewReader(signature))
```

caarlos0 (commented):
awesome work @erikgeiser, let me know if you need my help with anything! 🙏

caarlos0:
> We should discuss the user interface (config options) for this and whether or not we would also want to implement the option to sign with the `dpkg-sig` scheme.

I would vote to keep it simple and have only one implementation using only the origin role... if this happens to receive a lot of requests we can always add it later 🤔

apk/apk.go
erikgeiser: I wanted the user to know why I am parsing the maintainer email in the first place. I just spotted a type though (`send` instead of `set`)

caarlos0:
> The build is currently failing because I did not push my `go.mod` file which contains a redirection to a local checkout of the signing PR branch of `rpmpack`.

if you need we can fork it into the goreleaser org until it gets merged into the main repo...

erikgeiser:
The `apk` signature implementation from the original `apk` support PR was completely broken. That took me a while to figure out but I got it working now.

> I would vote to keep it simple and have only one implementation using only the origin role... if this happens to receive a lot of requests we can always add it later 🤔

Oops, I just added role configuration

> if you need we can fork it into the goreleaser org until it gets merged into the main repo...

Yeah, could you do that? Then I'll push a working build

What about `dpkg-sigs`?

caarlos0:
> Yeah, could you do that? Then I'll push a working build

done.

> Oops, I just added role configuration

oh, no problem then I guess :) 

> What about `dpkg-sigs`?

I would say to keep only a single implementation for now... keeping it simple and all :)

erikgeiser:
The RPM fork does not seem to have the necessary PR merged in yet I think, or am I missing something?

apk/apk.go
erikgeiser: fixed

apk/apk.go
erikgeiser: ;)

erikgeiser:
I'm not sure if we should wrap the signature relate config options in another `struct`. What do you think about the way the passphrase is read from the environment with a global fallback (so you don't have to set 3 vars, especially when RPM and DEB keys will normally be identical)?

caarlos0:
> I'm not sure if we should wrap the signature relate config options in another `struct`. What do you think about the way the passphrase is read from the environment with a global fallback (so you don't have to set 3 vars, especially when RPM and DEB keys will normally be identical)?

Yeah, sounds good to me :)



> The RPM fork does not seem to have the necessary PR merged in yet I think, or am I missing something? EDIT: NVM, merged it.

sorry I'm late for this, thanks

erikgeiser:
I was a little mad about `apk` signing `SHA1` hashes which have been proven to be broken. In theory you could take a signed `apk`, modify the packaged data, update the hash in `.PKGINFO` and brute force some metadata field until the `SHA1` hash collides with the original.

I then found [this mailing list entry](https://lists.alpinelinux.org/~alpine/devel/%3C257B6969-21FD-4D51-A8EC-95CB95CEF365%40ferrisellis.com%3E#%3C20180320150651.49fbcbae@vostro%3E) that says that `apk` does indeed support signatures over `SHA256` hashes, you just have to prefix the file name with `.SIGN.RSA256` instead of `.SIGN.RSA`. This did indeed work for the signature verification `apk verify foo.apk` but now `apk add` segfaults.

I also found an [IRC dump from earlier this year](https://irclogs.alpinelinux.org/%23alpine-devel-2020-02.log) where someone also noticed the segfaults with SHA256 hashes:

> \<reidrankin\> I say probably rsa, but it supports RSA (over sha-1), DSA, and sort-of RSA256 (using sha-256; unfortunately it segfaults because of a bug when trying to install these though)

erikgeiser:
@caarlos0 I removed the WIP, this is quite a sizeable chunk of new code, feel free to critique it as harsh as possible ;)

caarlos0:
many thanks for the hard work on this @erikgeiser, I will reserve some time to take a look at this tonight 🤘

deb/deb.go
```diff
@@ -80,27 +81,60 @@ func (*Deb) Package(info *nfpm.Info, deb io.Writer) (err error) {
...
 	}
+
+	if info.Deb.Signature.KeyFile != "" {
+		data := io.MultiReader(bytes.NewReader(debianBinary), bytes.NewReader(controlTarGz),
+			bytes.NewReader(dataTarGz))
+
+		sig, err := sign.PGPArmoredDetachSign(data, info.Deb.Signature.KeyFile,
+			info.Deb.Signature.KeyPassphrase)
+		if err != nil {
+			return errors.Wrap(err, "create signature")
```
caarlos0, line 121: I would put all these errors as exported errors, so API users can check against them...

internal/sign/rsa.go
```diff
@@ -0,0 +1,115 @@
...
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+)
+
+// RSASignSHA1Digest signs the provided SHA1 message digest. The key file
+// must be in the PEM format and can either be encrypted or not.
+func RSASignSHA1Digest(sha1Digest []byte, keyFile, passphrase string) ([]byte, error) {
+	if len(sha1Digest) != sha1.Size {
+		return nil, errors.New("digest is not a SHA256 hash")
```
caarlos0, line 20: sha256 or sha1 🤔

internal/sign/rsa.go
```diff
@@ -0,0 +1,115 @@
...
+
+	block, _ := pem.Decode(keyFileContent)
+	if block == nil {
+		return nil, errors.New("parse PEM block with private key")
+	}
+
+	blockData := block.Bytes
+	if x509.IsEncryptedPEMBlock(block) {
+		if passphrase == "" {
+			return nil, errors.New("key is encrypted no passphrase provided")
```
caarlos0, line 36: I would make the error exported, and also I think the wording is a bit off.. maybe `key is encrypted but no passphrase was provided`?

internal/sign/rsa.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return RSASignSHA1Digest(sha256Hash.Sum(nil), keyFile, passphrase)
+}
+
+// RSAVerifySHA1Digest is exported for use in tests and verifies a signature over the
+// provided SHA1 hash of a message. The key file must be in the PEM format.
+func RSAVerifySHA1Digest(sha1Digest, signature []byte, publicKeyFile string) error {
+	if len(sha1Digest) != sha1.Size {
+		return errors.New("digest is not a SHA256 hash")
```
caarlos0, line 76: also export this error..

heer also comparing sha1 and error says sha256...

internal/sign/rsa.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	keyFileContent, err := ioutil.ReadFile(publicKeyFile)
+	if err != nil {
+		return errors.Wrap(err, "reading key file")
+	}
+
+	block, _ := pem.Decode(keyFileContent)
+	if block == nil {
+		return errors.New("parse PEM block with public key")
```
caarlos0, line 86: export..

internal/sign/rsa.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return errors.Wrap(err, "parse PKIX public key")
```
caarlos0, line 91: export..

caarlos0 (commented):
Haven't finished yet, will take another look later.

Thanks again for the work 🙏

deb/deb.go
djgilcrease: Also you do not need to use `errors.Wrap` any more as that is built into go 1.13 where you can use `%w` https://blog.golang.org/go1.13-errors

Though it might be best to leave as is for now and do an explicit PR for removing `errors.Wrap`

internal/sign/rsa.go
erikgeiser: Good catch! This was a leftover from when I tried using `SHA256` hashes (see my comment from yesterday below the PR).

internal/sign/rsa.go
erikgeiser: Yes, of course. The missing `but` was a typo.

internal/sign/rsa.go
erikgeiser: Also an accidental leftover from trying to make it work with `SHA256`

deb/deb.go
erikgeiser: > I would put all these errors as exported errors, so API users can check against them...

Where would I export them? This is an internal package that should not be imported. Also, I would still wrap the underlying errors in the newly exported ones.

> Also you do not need to use errors.Wrap any more as that is built into go 1.13 where you can use %w https://blog.golang.org/go1.13-errors
>
> Though it might be best to leave as is for now and do an explicit PR for removing errors.Wrap

I agree, we should change wrapping over the whole codebase at some point. When I started working on `nfpm` `golangci-lint` would not allow me to use `%w`, but I just tried it again and now it works. Maybe I was using an outdated `golangci-lint` version from `brew` before or something.

Anyway, I'm a big proponent of `%w` and I already combed through some large codebases of mine to replace every `%v` in errors with `%w` 😄

nfpm.go
```diff
@@ -160,14 +181,42 @@ type RPM struct {
...
+	Triggers        DebTriggers  `yaml:"triggers,omitempty"`
+	Breaks          []string     `yaml:"breaks,omitempty"`
+	VersionMetadata string       `yaml:"metadata,omitempty"` // Deprecated: Moved to Info
+	Signature       DebSignature `yaml:"signature,omitempty"`
+}
+
+type DebSignature struct {
+	// PGP secret key, can be ASCII-armored
+	KeyFile       string `yaml:"key_file,omitempty"`
+	KeyPassphrase string // populated from environment variable
```
caarlos0, line 217: maybe 
```suggestion
	KeyPassphrase string `yaml:"-"` // populated from environment variable
```
to make sure its not populated from yaml?

nfpm.go
```diff
@@ -160,14 +181,42 @@ type RPM struct {
...
+}
+
+type APK struct {
+	Signature APKSignature `yaml:"signature,omitempty"`
+}
+
+type APKSignature struct {
+	// RSA private key in PEM format
+	KeyFile       string `yaml:"key_file,omitempty"`
+	KeyPassphrase string // populated from environment variable
```
caarlos0, line 200: maybe prevent it to be parsed from yaml:

```suggestion
	KeyPassphrase string `yaml:"-"` // populated from environment variable
```

nfpm.go
```diff
@@ -160,14 +181,42 @@ type RPM struct {
 	Compression string `yaml:"compression,omitempty"`
 	// https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html
 	ConfigNoReplaceFiles map[string]string `yaml:"config_noreplace_files,omitempty"`
+	Signature            RPMSignature      `yaml:"signature,omitempty"`
+}
+
+type RPMSignature struct {
+	// PGP secret key, can be ASCII-armored
+	KeyFile       string `yaml:"key_file,omitempty"`
+	KeyPassphrase string // populated from environment variable
```
caarlos0, line 190: maybe prevent it to be parsed from yaml:

```suggestion
	KeyPassphrase string `yaml:"-"` // populated from environment variable
```

nfpm_test.go
```diff
@@ -149,6 +150,46 @@ func TestParseFile(t *testing.T) {
...
+	const (
+		globalPass = "hunter2"
+		debPass    = "password123"
+		rpmPass    = "secret"
+		apkPass    = "foobar"
+		release    = "3"
+		version    = "1.0.0"
+	)
+
+	os.Clearenv()
```
caarlos0, line 163: seems like several tests in the same test method, maybe make it clearer using `t.Run()`? 🤔

deb/deb.go
caarlos0: > Where would I export them? This is an internal package that should not be imported. Also, I would still wrap the underlying errors in the newly exported ones.

`deb` is not an internal package... I would export them on the deb.go file 🤔

----

agree, lets migrate from errors.wrap in the near future 🙏

nfpm.go
erikgeiser: Oh, yes I forgot the ignore tag

nfpm_test.go
erikgeiser: Good idea

deb/deb.go
erikgeiser: > deb is not an internal package... I would export them on the deb.go file 🤔

You are right, but the other errors in your other comments are from an internal package

internal/sign/rsa.go
erikgeiser: From where? `rsa.go` is internal

internal/sign/rsa.go
erikgeiser: From where? `rsa.go` is internal

deb/deb.go
caarlos0: maybe just one, and wrap them? Maybe even on the main package (e.g. `nfpm.SigningError`) and then wrap it here 🤔

internal/sign/rsa.go
caarlos0: oh, indeed, if we don't need to check for specific errors on the implementions, we are safe leave as is.

deb/deb.go
erikgeiser: Do I understand you correctly that I should wrap the primitive errors thrown by `internal/sign` in an `nfpm.SigningError` and return that?

caarlos0: yeah, that's what I meant... but tbh, maybe in another PR? not sure if this will ever be needed... 🤔

erikgeiser: Like this?

```go
type SigningError struct {
	error
}

func (s *SigningError) String() string {
	return fmt.Sprintf("signing error: %v", s.error)
}

func (s *SigningError) Unwarp() error {
	return s.error
}

func (s *SigningError) Is(target error) bool {
	_, ok := target.(*SigningError)
	return ok
}```

**Comment by erikgeiser** [comment]
It looks like https://github.com/google/rpmpack/pull/39 is about to get merged

File: deb/deb.go (hunk 4 above)
Comment (erikgeiser) [reply deb/deb.go:121]: Hmm, I think we should really only add it as soon as someone needs this...

Comment (caarlos0) [reply deb/deb.go:121]: I would say just:

```go
var ErrSigning = errors.New("signing errors")
```

and then we can wrap errors as we already do (and later migrate to the new go way) and everything should work.

also your `Is` implementation is not correct I think... it should check for error equality, not type (yeah, its confusing). For type check we should use `errors.As` 🤷‍♂️

erikgeiser: But that would mean discarding the underlying error which is very useful to figure out what actually went wrong. Or did I miss something?

caarlos0: using the `fmt.Errorf("something: %w")` you can call `errors.Unwrap` to get the `SigningError` and also check type with `errors.As` (or, since this a "simple" error (not a `struct` that implements the error interface), `errors.Is` should work as well with no additional changes).

so, tl;dr, the caller could do:

```go
if errors.Is(err, nfpm.SigningError) {
```

and everything should work as expected. 🤔

erikgeiser: Sure but if `ErrSigning` is constructed with `errors.New` it will always be a "leaf" error, that cannot be unwrapped further. So everything that goes wrong in the signing process will only be reported as the general `ErrSigning`. Then the users won't know if the key file name is misspelled or the password is unset or wrong.

erikgeiser: This is why I wanted to implement `ErrSinging` in a way that it still holds the unwrapable underlying error thrown by the `internal/sign` package.

caarlos0: hmm, yeah, that makes sense... 

let's implement your idea then.

I think you don't need the `Is` method though, only `Unwrap` :)

caarlos0: (and `Error` of course)

erikgeiser: Yep, I'll do it tomorrow.

nfpm.go
```diff
@@ -322,3 +322,15 @@ func (info *Info) GetChangeLog() (log *chglog.PackageChangeLog, err error) {
 		Entries: entries,
 	}, nil
 }
+
+type ErrSigningFailure struct {
```
caarlos0, line 330: maybe just `ErrSign`, as err already implies a failure 🤔

erikgeiser: I actually thought quite a bit about the name. It seems like most packages name the error (the part after `Err`) to specify what went wrong or rather what type of error it was. It felt to me  `Sign` or `Signature` describes the result of a successful action instead of something that went wrong. And because I couldn't be more specific I settled on a `SignatureFailure`.

But actually I'm overthinking it majorly 😆
It should be your call though, so what do we do?

caarlos0: hmm, dunno either... you argument makes sense though... lets leave as is.

caarlos0:
IMHO good to merge, ccing @djgilcrease in case they want to take another look.

djgilcrease:
LGTM

caarlos0:

Issue #229, opened 2020-09-21
Increase test coverage


caarlos0:
We have a lot of code that's not very well tested, either by unit or integration tests.

https://codecov.io/gh/goreleaser/nfpm

It would be nice to improve that a bit 🙏

blueworrybear:
Hi @caarlos0!

I would like to contribute some test cases to increate the test coverage!

Please refer to [my commit](https://github.com/blueworrybear/nfpm/commit/15956c17c19efc159adbc3337b5c9f536fce8e3c) for the codes.

And here is the coverage report from [covergates](https://covergates.com/):

[![treemap](https://covergates.com/api/v1/reports/bu19l2a23akg00ckoqcg/treemap/?base=master&)](https://covergates.com/report/github/blueworrybear/nfpm?ref=15956c17c19efc159adbc3337b5c9f536fce8e3c)


### [Coverage: 80.1%](https://covergates.com/report/github/blueworrybear/nfpm?ref=15956c17c19efc159adbc3337b5c9f536fce8e3c)

||File|Coverage|
|--|--|--------|
||deb/deb.go|0.76|
||rpm/rpm.go|0.78|
||internal/sign/pgp.go|0.82|
|:arrow_up_small:|internal/sign/rsa.go|0.72|
||internal/glob/glob.go|0.94|
||nfpm.go|0.75|
||internal/files/files.go|0.83|
|:arrow_up_small:|apk/apk.go|0.80|

---

May I create a PR? Thank you and happy hacktoberfest!

caarlos0:
Yes, thanks 🙏🙏

erikgeiser:
Thanks for your contribution @blueworrybear. Sorry, but I don't really understand what the `sign.go` code is actually testing. `RSAVerifySHA1Digest` is actually not directly used by `nfpm` but exported to be used in tests in other packages. As far as I currently understand it, it is not really necessary to pull in all the dependencies to test different types of keys. I might be wrong, though.

Dean-Coakley:
@caarlos0 Is this issue still relevant?

I just had a look at master and the coverage looks pretty solid to me.
```
ok  	github.com/goreleaser/nfpm/v2	4.114s	coverage: 85.9% of statements
ok  	github.com/goreleaser/nfpm/v2/apk	4.233s	coverage: 81.7% of statements
?   	github.com/goreleaser/nfpm/v2/cmd/nfpm	[no test files]
[...]

PR #233, opened 2020-10-12
test: Add APK and RSA sign test

blueworrybear:
To increase test coverage.

See #229


erikgeiser:
Thanks, let me just copy my remark from issue #229 to discuss it here: 

Thanks for your contribution @blueworrybear. Sorry, but I don't really understand what the sign.go code is actually testing. RSAVerifySHA1Digest is actually not directly used by nfpm but exported to be used in tests in other packages. As far as I currently understand it, it is not really necessary to pull in all the dependencies to test different types of keys. I might be wrong, though.

internal/sign/rsa_test.go
```diff
@@ -49,3 +60,51 @@ func TestInvalidHash(t *testing.T) {
...
+		{
+			Type:  "PUBLIC KEY",
+			Bytes: x509Bytes,
+		},
+	}
+
+	for _, pemKey := range pemKeys {
+		cwd, _ := os.Getwd()
+
+		keyFile, err := ioutil.TempFile(cwd, "*.pem")
```
erikgeiser, line 98: Maybe just create the temp file in the OS temp folder (`ioutil.TempFile("", "*.pem")`). This way the working tree is not polluted if the test is interrupted by CTRL+C in the wrong time.

blueworrybear:
> Thanks, let me just copy my remark from issue #229 to discuss it here:
> 
> Thanks for your contribution @blueworrybear. Sorry, but I don't really understand what the sign.go code is actually testing. RSAVerifySHA1Digest is actually not directly used by nfpm but exported to be used in tests in other packages. As far as I currently understand it, it is not really necessary to pull in all the dependencies to test different types of keys. I might be wrong, though.

I see, I misunderstand the comment of `RSAVerifySHA1Digest`. Since coverage report counts in `RSAVerifySHA1Digest`, should I keep the tests or remove it?

internal/sign/rsa_test.go
blueworrybear: fixed!

erikgeiser:
The function was added to verify APK signatures which are not that well documented. As far as the documentation goes, only RSA keys are supported. So I don't know if ECDSA does actually even work with APKs. In my very limited understanding of the crypto involved, I think ECDSA is also one of the worse elliptic curve schemes. If I'm correct the ECDSA key test case tests a scenario that does not happen in practice.

What I don't understand is that the ECDSA works at all given that I did a type assertion that should require the key to be an RSA key:

https://github.com/goreleaser/nfpm/blob/cfc81695484e824e305f83c83ff81e1bcc3a69c9/internal/sign/rsa.go#L98-L101

Testing the function even though it is not used in the actual production code is still not a bad idea but if you want you can remove the key creation and just test with the keys in the `testdata` directory that are used for the other tests.

caarlos0:
> @caarlos0 wdyt?

sgtm

(sorry I'm late, notification got lost in a sea of notifications_

erikgeiser:
I rebased this PR to `master` and replaced the RSA tests with detailed tests for different error conditions. It can be merged now if CI succeeds.

caarlos0:
Awesome, thanks

PR #234, opened 2020-10-21
deps: Update zglob.

rtpt-erikgeiser:
This PR updates `go-zglob` to the first commit that supports escaping. This allows `nfpm` to package files that contain special glob characters such as `{` and `}`.


rtpt-erikgeiser:
CI fails because fedora package servers are down and the (unrelated) metapackage test times out while installing `zsh` and `fish`.

caarlos0:
Thanks!

PR #238, opened 2020-10-30
fix: Fix hash name typo.

### LINKED ISSUE #237: Use SHA256 or rename to SHA1
https://github.com/goreleaser/nfpm/blob/3d7580ade2e01168e7a3487c145d44b98bc1eb3a/internal/sign/rsa.go#L67

This line is confusing, you talk about a SHA256 hash but are actually using SHA1.

erikgeiser:
Closes #237.


caarlos0:
thanks

PR #240, opened 2020-11-02
feat: add support for deb templates

vandot:
Add support for `templates` file inside `DEBIAN` that is needed for `debconf` to be able to ask question user during installation. More on subject [here](http://www.fifi.org/doc/debconf-doc/tutorial.html) and [here](https://manpages.debian.org/testing/debconf-doc/debconf-devel.7.en.html)

### DIFF
--- a/nfpm.go
+++ b/nfpm.go
@@ -204,7 +204,6 @@ type APKSignature struct {
 // Deb is custom configs that are only available on deb packages.
 type Deb struct {
 	Scripts         DebScripts   `yaml:"scripts,omitempty"`
-	Templates       string       `yaml:"templates,omitempty"`
 	Triggers        DebTriggers  `yaml:"triggers,omitempty"`
 	Breaks          []string     `yaml:"breaks,omitempty"`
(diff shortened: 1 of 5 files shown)


djgilcrease:
Is there a reason this cannot just be handled with the normal files process?

vandot:
@djgilcrease `templates` file needs to be inside `DEBIAN` folder, maybe I missed something but with `files` and `conffiles` files are placed in relative installation path

nfpm.go
```diff
@@ -204,6 +204,7 @@ type APKSignature struct {
 // Deb is custom configs that are only available on deb packages.
 type Deb struct {
 	Scripts         DebScripts   `yaml:"scripts,omitempty"`
+	Templates       string       `yaml:"templates,omitempty"`
```
djgilcrease, line 207: I would move this into the `DebScripts` object

djgilcrease (approved):
LGTM

caarlos0 (approved):
can you add it to the docs too?

other than that LGTM

vandot:
Done, added to the docs.

PR #244, opened 2020-11-05
feat: replace glob library

### LINKED ISSUE #232: Globbing Replacement
As far as I am aware, it is currently impossible to package files or directories that are enclosed in curly braces (such as `{foo}`). An example would be Firefox extensions which use `{UUID}` directories. The reason for this is that https://github.com/mattn/go-zglob does not seem to support wildcard escaping and because also because of a bug (https://github.com/mattn/go-zglob/issues/32).

A possible alternative to replace https://github.com/mattn/go-zglob with is https://github.com/gobwas/glob which supports the same features, escaping and more extensive tests. However, this library only does string matching and the directory walking will have to be performed in `nfpm`.

caarlos0:
BREAKING CHANGE: may have unintended side-effects.

closes #232

error message when no files match a glog changed, because now its handled in another place... but seems like tests are still passing...

### DIFF
--- a/acceptance/testdata/complex.386.yaml
+++ b/acceptance/testdata/complex.386.yaml
@@ -19,7 +19,7 @@ homepage: "https://foobar.org"
 license: "MIT"
 files:
   ../testdata/fake: "/usr/local/bin/fake"
-  ./testdata/folder/**/*: "/usr/share/whatever/folder/"
+  ./testdata/folder/*: "/usr/share/whatever/folder/"
 config_files:
   ../testdata/whatever.conf: "/etc/foo/whatever.conf"
 empty_folders:
--- a/apk/apk_test.go
+++ b/apk/apk_test.go
@@ -168,7 +168,7 @@ func TestFileDoesNotExist(t *testing.T) {
 					"bash",
 				},
 				Files: map[string]string{
-					"../testdata/": "/usr/local/bin/fake",
+					"../testdata/fake": "/usr/local/bin/fake",
 				},
 				ConfigFiles: map[string]string{
 					"../testdata/whatever.confzzz": "/etc/fake/fake.conf",
@@ -177,7 +177,7 @@ func TestFileDoesNotExist(t *testing.T) {
 		}),
 		ioutil.Discard,
 	)
-	assert.EqualError(t, err, "glob failed: ../testdata/whatever.confzzz: file does not exist")
+	assert.EqualError(t, err, "glob failed: ../testdata/whatever.confzzz: no matching files")
 }
 
 func TestNoFiles(t *testing.T) {
--- a/acceptance/testdata/complex.yaml
+++ b/acceptance/testdata/complex.yaml
@@ -21,7 +21,7 @@ homepage: "https://foobar.org"
 license: "MIT"
 files:
   ../testdata/fake: "/usr/local/bin/fake"
-  ./testdata/folder/**/*: "/usr/share/whatever/folder/"
+  ./testdata/folder/*: "/usr/share/whatever/folder/"
 config_files:
   ../testdata/whatever.conf: "/etc/foo/whatever.conf"
 empty_folders:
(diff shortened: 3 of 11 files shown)


erikgeiser:
Is it really correct to start walking at the working directory? Shouldn't it start on the path in the pattern before the first wildcard?

caarlos0:
> Is it really correct to start walking at the working directory? Shouldn't it start on the path in the pattern before the first wildcard?

that wouldn't work if the first part of the path has a wildcard... 🤔

erikgeiser:
In this case it should indeed start in the working directory but always starting in the working directory does not work for patterns with an absolute path.

caarlos0:
true, but I'm not sure they work with zglob either 🤔 

maybe we should always use absolute paths? (e.g. `filepath.Abs(".")` and then `g.Match(filepath.Abs(path))`)

erikgeiser:
If I understand correctly that does not work if the user sets an absolute path which is completely different from the current directory. An example would be if a user pulls the files from a network share or something like that.

caarlos0:
I think doesn't work in the current version either, as zglob assumes `.` as start point too AFAIK 🤔

caarlos0:
I'm thinking in maybe create a separated lib, copying the tests from zglob et al, to make sure we don't break anything...

Will do that later today.

erikgeiser:
Nope, `zglob` determines the "root" (`zenv.root` in `zglob.go`) and walks from threre. This means that `/foo/bar/**/baz` would start walking in `/foo/bar`, regardless of the current directory. For relative paths, `zenv.root` is `.`.

Also, in `zglob` directories are skipped when nothing inside them would match anyway. This means that for the pattern `/foo/bar/x*/*`, `zglob` would start walking in `/foo/bar` but it would skip directories like `/foo/bar/baz` altogether.

caarlos0:
Yeah, was looking into the source as well, you're right.

Maybe we can fork zglob and just replace the glob part and keep the rest. Will look into that

erikgeiser:
It is probably not that easy because the walk function depends on the information gathered during pattern parsing. I've read the code multiple times and I don't understand the glob parsing at all, so I'm not sure if the directory skipping I mentioned before does work or not.

Maybe the best approach is to find a prefix of the pattern that does not contain wildcards and walk from there. Maybe the subpackages of the new glob libraries can help with that.

erikgeiser:
```go
rootNode, err := ast.Parse(lexer.NewLexer("/foo/bar/**/ba{z,y}"))
if err != nil {
	die(err)
}

fmt.Println(rootNode.Children[0].Value.(ast.Text).Text) # /foo/bar
```

For reference, the AST in this case looks like this:
```
Pattern (<nil>)
  Text ({/foo/bar/})
  Super (<nil>)
  Text ({/ba})
  AnyOf (<nil>)
    Pattern (<nil>)
      Text ({z})
    Pattern (<nil>)
      Text ({y})
```

So if the first child node of the pattern is text, we can use it as the starting point. Otherwise we can start at `.`.

erikgeiser:
If you want I can also work on this. If you want you can shoot me a message at the Gophers slack to coordinate our effort.

acceptance/testdata/complex.386.yaml
```diff
@@ -19,7 +19,7 @@ homepage: "https://foobar.org"
 license: "MIT"
 files:
   ../testdata/fake: "/usr/local/bin/fake"
-  ./testdata/folder/**/*: "/usr/share/whatever/folder/"
```
caarlos0, line 22: I'm not sure which is wrong, or if any of them are, but results are different now 🤔

apk/apk_test.go
```diff
@@ -168,7 +168,7 @@ func TestFileDoesNotExist(t *testing.T) {
 					"bash",
 				},
 				Files: map[string]string{
-					"../testdata/": "/usr/local/bin/fake",
+					"../testdata/fake": "/usr/local/bin/fake",
```
caarlos0, line 171: this should have been failing before, wasn't 🤣

erikgeiser (approved):

erikgeiser:
lgtm, but can I add `DisableGlobbing` option to this PR before we merge?

caarlos0:
> lgtm, but can I add `DisableGlobbing` option to this PR before we merge?

let's do it in another PR, will wait for it before releasing... :)

PR #51, opened 2019-04-10
issue46: Add support for a Release flag

jimrobinson:
This patch adds support for specifying an rpm Release flag other than the default '1'.

```
$ head -5 nfpm.yaml
name: "csync"
arch: "amd64"
platform: "linux"
version: "1.0.0"
release: "2"

$ make rpm
nfpm pkg --target csync-"1.0.0"-"2".x86_64.rpm
using rpm packager...
created package: csync-1.0.0-2.x86_64.rpm

$ rpm -qip ./csync-1.0.0-2.x86_64.rpm | head -5
Name        : csync
Version     : 1.0.0
Release     : 2
Architecture: x86_64
Install Date: (not installed)
```

PR #245, opened 2020-11-07
feat: Support RPM ghost files

wwade:
From the [Maximum RPM] guide:

    The %ghost Directive

    As we mentioned in the Section called The %files List, if a file
    is specified in the %files list, that file will automatically be
    included in the package. There are times when a file should be
    owned by the package but not installed - log files and state files
    are good examples of cases you might desire this to happen.

    The way to achieve this, is to use the %ghost directive. By adding
    this directive to the line containing a file, RPM will know about
    the ghosted file, but will not add it to the package.

Ghost files are specified using `Type: rpmpack.GhostFile` when
constructing the output package.

The test ensures that we have:
 - The target file present in the RPM.
 - The specified mode attributes (although not really important,
 either).
 - An empty file *[1]* in the RPM for the named ghost, 
see [google/rpmpack #51].

*[1] Instead of an empty file, no file should be created, but this is
not possible until the upstream issue is resolved.*

[Maximum RPM]:
http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html

[google/rpmpack #51]: https://github.com/google/rpmpack/issues/51

### DIFF
--- a/nfpm.go
+++ b/nfpm.go
@@ -182,6 +182,7 @@ type RPM struct {
 	// https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html
 	ConfigNoReplaceFiles map[string]string `yaml:"config_noreplace_files,omitempty"`
 	Signature            RPMSignature      `yaml:"signature,omitempty"`
+	GhostFiles           []string          `yaml:"ghost_files,omitempty"`
 }
 
 type RPMSignature struct {
--- a/rpm/rpm.go
+++ b/rpm/rpm.go
@@ -350,6 +350,19 @@ func createFilesInsideRPM(info *nfpm.Info, rpm *rpmpack.RPM) error {
 		}
 	}
 
+	// note: the ghost files will be created as empty files when the package is installed, which is not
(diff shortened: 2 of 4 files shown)


nfpm.go
```diff
@@ -181,6 +181,7 @@ type RPM struct {
 	// https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html
 	ConfigNoReplaceFiles map[string]string `yaml:"config_noreplace_files,omitempty"`
 	Signature            RPMSignature      `yaml:"signature,omitempty"`
+	Ghosts               []string          `yaml:"ghosts,omitempty"`
```
caarlos0, line 184: maybe GhostFiles instead (to go with ConfigNoReplaceFiles)?

```suggestion
	GhostFiles          []string          `yaml:"ghost_files,omitempty"`
```

wwade: Sounds good, thanks for reviewing.  I'll make the changes.

caarlos0 (approved):

caarlos0:
Thanks!

Issue #252, opened 2020-11-10
can't get module


caarlos0:
not sure what's happening, but if I try to go get latest, I get:

```
~/Code/goreleaser/goreleaser master
❯ go get github.com/goreleaser/nfpm@v1.10.0
go get: github.com/goreleaser/nfpm@v1.10.0/go.mod: verifying module: github.com/goreleaser/nfpm@v1.10.0/go.mod: reading https://sum.golang.org/lookup/github.com/goreleaser/nfpm@v1.10.0: 410 Gone
	server response: not found: create zip: malformed file path "testdata/{file}*": invalid char '*'
```

even tried to create a .1 tag, thinking maybe some weird caching somewhere somehow (after deleting my local cache), and it still fails:

```
~/Code/goreleaser/goreleaser master 20s
❯ go get github.com/goreleaser/nfpm@v1.10.1
go get: github.com/goreleaser/nfpm@v1.10.1/go.mod: verifying module: github.com/goreleaser/nfpm@v1.10.1/go.mod: reading https://sum.golang.org/lookup/github.com/goreleaser/nfpm@v1.10.1: 410 Gone
	server response: not found: create zip: malformed file path "testdata/{file}*": invalid char '*'
```

I have no idea why.

caarlos0:
oh, actually I do, reading the error helps: invalid file name: `testdata/{file}` 🤣

caarlos0:
fixed, closing

PR #253, opened 2020-11-10
fix: invalid zip file

caarlos0:
refs #252


caarlos0:
Confirmed this works with:

```
go mod edit -replace github.com/goreleaser/nfpm=github.com/goreleaser/nfpm@fix
go mod tidy
```

Issue #43, opened 2018-11-28
Provide files as tgz


tcurdt:
It seems like right now there is no way to specify a full tgz for the files that should go into the package. And there is no way to specify the ownership and permissions in detail (that I see).

Would be great if one could provide a complete tgz as the data section of package.

tcurdt:
That sounds very sceptical :) May I ask why?
How do you setup permissions of the files included in the package at the moment instead?

caarlos0:
sorry, didn't meant to sound very skeptical haha 

right now, as far as I remember, it uses whatever permissions the files already have, although I agree that's not ideal.

I use it only on goreleaser itself, and for what goreleaser supports that was never a problem, but to support "power users" I agree there must be a better way...

tcurdt:
If you have any thoughts on how you this could or should be added best I am all ears.
Maybe I can then work on a PR.

jarondl:
Now with #79 by @djgilcrease , nfpm can perhaps offer using `rpmpack.FromTar` as an additional input method.

djgilcrease:
I would say add a `FilesTar string` to [Overridables](https://github.com/goreleaser/nfpm/blob/master/nfpm.go#L136) make sure it is mutually exclusive with the `Files []string` then update [createDataTarGz](https://github.com/goreleaser/nfpm/blob/master/deb/deb.go#L91) to use the passed in `FilesTar` if set && update [Package](https://github.com/goreleaser/nfpm/blob/master/rpm/rpm.go#L45) to use [FromTar](https://github.com/google/rpmpack/blob/master/tar.go#L28) if `FilesTar` is set

mig4:
Yeah, I just wanted to confirm the expectation was that this would be an additional option in the config file, instead of a CLI arg for example. As for in addition/instead, whatever makes most sense. I'll give it a spin and propose something and we can discuss the details there sure.

Many thanks for the pointers @djgilcrease :grin:

caarlos0:
closing due to lack of interest from the community

tcurdt:
@caarlos0 is there any other way to specify ownership and permissions of files inside the package by now?

caarlos0:
yes, https://nfpm.goreleaser.com/configuration/?h=file_info#reference

PR #255, opened 2020-11-15
feat: try to clean up and simplify the file adding interface

djgilcrease:
The goal is to make this backward compatible with current config files so people can stay with the simplified format until they need more advanced features. 

Internally all packagers `ONLY` use the Contents data which is a merging of all the old file specifications into one list. IF a file was specified in a packager specific override (eg `rpm.Files`) then the `content.Packager` field is set to the correct packager automatically which ensures files for other packagers do not get mixed together.

Example
```yaml
name: "contents foo"
arch: "amd64"
version: "v1.2.3"
contents:
  - src: "./testdata/*.conf"
    dst: "/etc/foo/"
    type: "config|noreplace"
    file_info:
      mode: 0644
      mtime: 2008-01-02T15:04:05Z
      owner: me
      group: whatever
  - src: "./testdata/whatever2.conf"
    dst: "/etc/rpm/whatever.conf"
    type: "config|noreplace"
    packager: "rpm"
    file_info:
      mode: 0644
  - src: "./testdata/whatever2.conf"
    dst: "/etc/apk/whatever.conf"
    type: "config|noreplace"
    packager: "apk"
    file_info:
      mode: 0644
  - src: "./testdata/whatever2.conf"
    dst: "/etc/deb/whatever.conf"
    type: "config|noreplace"
    packager: "deb"
    file_info:
      mode: 0644
```

The `files`, `config_files`, `symlinks`, `rpm.config_noreplace_files`, and `rpm.ghost_files` are merged into `contents` and all packagers only the the data from `contents` now. This makes it simple for people to leave their config files as is and not break anything, but as they need more advanced features, like control over the file mode, owner, group, etc they can convert all their files to the contents as well.

I have also set it up so the simple mapping will work as well
```yaml
contents:
  a: b
  a2: b2
```

Also this is setup to be ready to handle contents as a string in prep for #43 but is not implemented
```yaml
contents: /path/to/a/tgz
```


erikgeiser:
This is a pretty significant change and should definitely be discussed. I do think your approach is cleaner and more elegant but the old one is probably more pragmatic, since most people probably have a whole bunch of regular files and rarely any special files. I have no data to back that up though. I generate my configs automatically as a part of a bigger custom build process, but for those writing configs by hand, it will be much more to type.

I'll have a look at the exact changes when I find some time, maybe tomorrow.

djgilcrease:
Ya, that's why I am making sure to ensure backwards compatibility so most people can continue to use the simple config format if they want.

erikgeiser:
If I read your code correctly you planned to implement a custom unmarshal method so `files` can be both a YAML mapping and a list of `FileToCopy` in the YAML file. This is a great way to implement it.

Maybe we can use `src` and `dst` as YAML keys to make it more convenient for regular files in the new format.

Also I dislike the name of the type `FileToCopy`. If it's a glob, it represents multiple files that need to be copied. If it's a ghost file, nothing is to be copied. Maybe we could rename it to `Content`.

internal/files/files.go
```diff
@@ -1,38 +1,78 @@
...
 			return nil, err
 		}
-		for src, dst := range globbed {
-			files = append(files, FileToCopy{src, dst})
+		appendAndSort(globbed, files)
+	}
+
+	return files, nil
+}
+// Expand gathers all of the real files to be copied into the package.
```
caarlos0, line 50: same godocs as Expand..

internal/files/files.go
```diff
@@ -1,38 +1,78 @@
...
-	Source      string
-	Destination string
+	Source string `yaml:"source,omitempty"`
+	Destination string `yaml:"destination,omitempty"`
+	Type string `yaml:"type,omitempty"`
+	Packager string `yaml:"packager,omitempty`
+	Mode uint `yaml:"mode,omitempty"`
 }
 
+//type tmpFruitBasket []map[string]yaml.Node
```
caarlos0, line 20: rm commented code

erikgeiser:
This is also nice for #43 because now we can then completely decide what do based on the YAML node type:
* **Mapping:** `src`->`dst` mapping of regular files, no non-regular files supported (only via depecated options), this is nice for very simple packages
* **Sequence:** The new rich file types you introduced here
* **String:** Path to a `.tar.gz` file

I think then `Content` or `Contents` would indeed be a much more fitting name.

internal/files/files.go
djgilcrease: removed the expand function

internal/files/files.go
djgilcrease: done

djgilcrease:
If we want a new field instead of trying to override the current `files` field that should actually be easier and cleaner than what I am having to do now I think. I can still have code that merges the `files`, `config_files`, `symlinks`, `rpm.config_noreplace_files`, and `rpm.ghost_files` into `contents` so all the packagers only use the `contents`

```yaml
contents:
  - src: "something"
    dst: "/opt/pkg/something"
    type: "config"
    packager: "rpm"
    mode: 0644
```

give me a day or so to play with that concept to see if it turns out cleaner

djgilcrease:
> give me a day or so to play with that concept to see if it turns out cleaner

Much cleaner code now that I am not trying to mess with existing fields

www/docs/configuration.md
```diff
@@ -85,31 +85,75 @@ suggests:
 # Packages it conflicts with. (overridable)
 conflicts:
   - mercurial
-
-# Files to add to the package. (overridable)
+    
+# Contents to att to the package
```
caarlos0, line 89:
```suggestion
# Contents to add to the package
```

caarlos0 (approved):
this looks awesome, thanks for working on this @djgilcrease ! 🚀

caarlos0:
> The goal is to make this backward compatible with current config files so people can stay with the simplified format until they need more advanced features.

I would say we deprecate the old format. The new format is simple enough for the basic use case as well, and then when users need to change it, its not that much of a change?

wdyt?

if you agree, we can copy the deprecate package from goreleaser and do basically the same thing here, and remove the old option after 6mo or so. (and we can make this all in another PR since this one is already pretty big)

PR #262, opened 2020-12-08
fix: ensure nfpm can run on windows and still build correct rpm, deb, and apk packages

djgilcrease:
- [x] Add testing on windows to the actions
- [x] Figure out why the rpm tests cannot remove the tmp rpm files created


caarlos0:
LGTM so far, once this is fixed/released, will do the same on goreleaser.

.github/workflows/build.yml
```diff
@@ -9,6 +9,27 @@ on:
...
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
```
caarlos0, line 22: should be 1.15?

```suggestion
          go-version: 1.15
```

go.mod
```diff
@@ -3,22 +3,44 @@ module github.com/goreleaser/nfpm
 go 1.14
 
 require (
+	github.com/Djarvur/go-err113 v0.1.0 // indirect
```
caarlos0, line 6: missing a `go mod tidy`?

internal/files/files.go
```diff
@@ -43,3 +44,9 @@ func Expand(filesSrcDstMap map[string]string, disableGlobbing bool) ([]FileToCop
 	})
 	return files, nil
 }
+
+// ToNixPath converts the path to the nix style path
+// Windows style path separators are escape characters so cause issues.
```
caarlos0, line 49:
```suggestion
// ToNixPath converts the given path to a nix-style path.
//
// Windows-style path separators are considered escape 
// characters by some libraries, which can cause issues.
```

caarlos0 (commented):
the changes in the www/docs folder seem unrelated 🤔

go.mod
djgilcrease: even after a `make deps` and an explicit `go mod tidy` this still shows up

caarlos0 (approved):

PR #267, opened 2020-12-22
feat: remove deprecated options

caarlos0:
regarding #255, removing the old options.

We will then version this as v2 (breaking change). Easier than keep a lot of things working.

Will still have to do the deprecation notice et all on goreleaser, but its on a single place then.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14
+          go-version: 1.15
       -
         name: Cache Go modules
         uses: actions/cache@v1
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ conduct](/CODE_OF_CONDUCT.md).
 Prerequisites:
 
 * `make`
-* [Go 1.14+](https://golang.org/doc/install)
+* [Go 1.15+](https://golang.org/doc/install)
 
 Clone `nfpm` from source:
 
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ setup:
 
 
 pull_test_imgs:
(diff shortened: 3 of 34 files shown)


caarlos0:
@djgilcrease I'm not sure why, but symlink tests on rpm and deb are failing 🤔 

I think it wasn't supposed to work before actually, since the actual file is never added to the package... does that makes sense?

djgilcrease:
As part of this make sure you read https://blog.golang.org/v2-go-modules many people at my work have tried to bump to v2 but not read that and broken things in odd ways.

caarlos0:
> As part of this make sure you read https://blog.golang.org/v2-go-modules many people at my work have tried to bump to v2 but not read that and broken things in odd ways.

yep, here's a picture of me reading it the last time I needed: 

image

😂

caarlos0:
> As part of this make sure you read https://blog.golang.org/v2-go-modules many people at my work have tried to bump to v2 but not read that and broken things in odd ways.

In the past, I simply changed the module to `/v2` in `go.mod` (e.g. https://github.com/caarlos0/env), but yeah, they recommend a `v2` subfolder, which I absolutely hate and 0/10 will not do it. Just renaming works for go mod and non-go mod usage, but old versions won't work under go mods anymore (if I recall correctly).

Maybe just keep a v2 branch? Anyone with more experience with that?

caarlos0:
> Other strategies may keep major versions on separate branches. However, if v2+ source code is on the repository's default branch (usually master), tools that are not version-aware — including the go command in GOPATH mode — may not distinguish between major versions.

IMHO ¯\_(ツ)_/¯ and we can just append `/v2` to the `go.mod` file... YOLO

djgilcrease:
> > Other strategies may keep major versions on separate branches. However, if v2+ source code is on the repository's default branch (usually master), tools that are not version-aware — including the go command in GOPATH mode — may not distinguish between major versions.
> 
> IMHO ¯_(ツ)_/¯ and we can just append `/v2` to the `go.mod` file... YOLO

Ya that works, I don't really like the branch or sub folder method either, but it does make supporting older versions easier if we need to for some reason.

djgilcrease (approved):
LGTM

PR #268, opened 2020-12-23
feat: version as v2

caarlos0:
following #267, version as v2 now.

BREAKING CHANGE: changes import path to add a /v2 sufffix


djgilcrease (approved):
Lgtm

PR #269, opened 2020-12-23
feat: more deprecated things to remove

caarlos0:
removed more deprecated things...

my plan is to roll it in v2.1 (because I forgot about it in v2.0)


djgilcrease (approved):
Good to go

PR #273, opened 2020-12-30
fix: race condition and remove unused code

caarlos0:
remove unused custom yaml parse and make some code a bit clearer...

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -50,25 +50,33 @@ func (c Contents) Less(i, j int) bool {
 	return a.Destination < b.Destination
 }
 
-func (c *Content) WithFileInfoDefaults() {
-	if c.FileInfo == nil {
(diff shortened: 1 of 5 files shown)


files/files.go
```diff
@@ -91,25 +50,33 @@ func (c Contents) Less(i, j int) bool {
 	return a.Destination < b.Destination
 }
 
-func (c *Content) WithFileInfoDefaults() {
-	if c.FileInfo == nil {
-		c.FileInfo = &ContentFileInfo{}
+func (c *Content) WithFileInfoDefaults() *Content {
```
caarlos0, line 53: this method could be called by multiple threads at once, therefore we probably need to modify and return a copy instead.

djgilcrease: 👍

djgilcrease (approved):
LGTM

Issue #276, opened 2020-12-29
NFPM cannot sign packages with only the subkey present


janoszen:
**Describe the bug**

When attempting to sign packages with subkeys only (no main key present) the signing fails with the following error message:

```
release failed after 11.33s error=nfpm failed: signing error: armored detach sign: decoding armored PGP keyring: openpgp: unsupported feature: hash for S2K function: 0
```

This is most likely due to golang/go#23226

**To Reproduce**

```yaml
builds:
nfpms:
  - rpm:
      signature:
        key_file: /tmp/signing.gpg
    deb:
      signature:
        key_file: /tmp/signing.gpg
    apk:
      signature:
        key_file: /tmp/signing.gpg
```

The key needs to be generated with the following steps:

```bash
$ gpg --full-generate-key
# Complete the key setup
$ gpg --edit-key test@example.com
gpg> addkey
# Complete adding the subkey
gpg> quit
Save changes? (y/N) y
$ gpg --list-keys --with-keygrip
$ rm -rf ~/.gnupg/private-keys-v1.d/ADD-MAIN-KEY-KEYGRIP-HERE.key
$ gpg -a --export-secret-subkey test@example.com >/tmp/signing.gpg
```

**Expected behavior**

The signing should either work with the GnuPG agent instead of using a file, or it should read the GnuPG key correctly.

**Environment (please complete the following information):**

- OS: linux
- OS version: Linux janoszen-desktop 4.19.128-janoszen-wsl goreleaser/goreleaser#2 SMP Tue Dec 15 12:51:43 CET 2020 x86_64 x86_64 x86_64 GNU/Linux
- GoReleaser Version 0.149.0

**Additional context:**

You can verify that the key is affected by this issue by running `gpg --list-packets KEYFILE.gpg` and looking for the `gnu-dummy S2K` packet.

caarlos0:
cc/ @erikgeiser can you take a look?

erikgeiser:
`hash for S2K function: 0` probably means that the key parser could not read the whole buffer (2 bytes) because hash IDs are >0 and the zero is probably the uninitialized buffer. So we can probably rule out that the problem is an unsupported algorithm. It does indeed seem that https://github.com/golang/go/issues/23226 is related but in that case, the **secret** key material is missing from a subkey.

[...]

PR #278, opened 2021-01-08
Adding test for issue #276 GPG subkeys without main keys

janoszen:
This PR adds a test for goreleaser/nfpm#276, but does not fix it.


caarlos0:
does this still makes sense?

djgilcrease:
No, I merged these tests into my fix

PR #283, opened 2021-01-16
feat: add a keep_version flag

wellenvogel:
I'm using nfpm for some of my projects. As I already started other projects that also produce packages with a slightly different versioning schema (using other tools) I would like to have a chance to keep my version completely unchanged from the config to the created package.
To be fully compatible I introduced a flag to switch off the semver handling.

As I encountered problems when afterwards building a docker image I added a staticbuild task to the Makefile.

Working image at https://hub.docker.com/repository/docker/wellenvogel/nfpm: wellenvogel/nfpm:1.0


djgilcrease (changes requested):
Looks good after a rebase

caarlos0:
hey, thanks for the pr and sorry for the delay to review, lost it in the notifications I think...

---

I would say to do something more like:

```
version_schema: semver (default)|none
```

so its hopefully clearer.

wdyt?

---

also need to mention the new option in the docs...

djgilcrease:
> ```
> version_schema: semver (default)|none
> ```

I like this, then I would change the check to a switch to make it easier to add additional version schema in the future.
```go
        switch info.VersionSchema {
            case "none":
                // No change to the version or prerelease info set in the YAML file
                break
            case "semver":
                fallthrough
            default:
                info.parseSemver()
        }
```

caarlos0:
> I like this, then I would change the check to a switch to make it easier to add additional version schema in the future.

yep :)

wellenvogel:
For me it also looks good.
Any solution that allows to set a plain version is completely ok.

djgilcrease (approved):

djgilcrease:
Superceeded by #324

PR #287, opened 2021-01-27
fix: set the alpine linux version to v3.12 so the acceptance tests pass

djgilcrease:
I will create another issue to track the requirement that we start embedding the checksums in the apk files.


caarlos0 (approved):

caarlos0:
Nice, thanks

PR #290, opened 2021-01-30
fix: Prevent destination collisions

erikgeiser:
This PR makes `ExpandContentGlobs` return an error when multiple contents share the same destination. Otherwise, the destination would just be overwritten.

This leads to particularly nasty broken packages when the symlink convention is misunderstood, so I improved the documentation for that too. If you think that doesn't belong in this PR I will split the PR. But first hear me out:

Everyone I know (including myself) always forgets the order of arguments in `ln -s file1 file2` such that most of the time `ln` calls are preceded by `man ln`. So refering to `ln` in the `nfpm` docs for symlinks might not be the best idea. I learned this the hard way when I added a symlink to a package thinking that

```
- src: /usr/bin/mybinary
  dst: /path/to/my/actual/binary
  type: symlink
```

means that I create a symlink that points from the source ` /usr/bin/mybinary` to the destination `/path/to/my/actual/binary`, but actually it is the other way around. That is definitely confusing but ultimately not a bad idea because this way "dst" always refers to the file created by the package. So this just needed more documentation.

But the end of the story was that my package resulted in a dangling symlink at `/path/to/my/actual/binary` which completely overwrote the actual binary 😄. So this PR addresses both issues that together lead to this broken result.

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -1,6 +1,7 @@
 package files
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
@@ -51,7 +52,7 @@ func (c Contents) Less(i, j int) bool {
 }
 
 func (c *Content) WithFileInfoDefaults() *Content {
-	var cc = &Content{
+	cc := &Content{
 		Source:      c.Source,
 		Destination: c.Destination,
 		Type:        c.Type,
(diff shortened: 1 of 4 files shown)


djgilcrease (approved):
LGTM

caarlos0 (approved):

PR #291, opened 2021-02-02
fix: add the PAXHeaders for file checksum required by alpine 3.13+

djgilcrease:
closes #288

PR #295, opened 2021-02-10
fix: apk checksums

caarlos0:
- looking at official packages, seems like only SHA1 is needed (made a small thing to check that: https://github.com/caarlos0/apkcat)
- added tests
- fixed the missing write to the hasher function

The packages actually had the file contents as a header as well as far as I understand. 

PS: tested it with goreleaser itself as well.

Refs https://github.com/goreleaser/nfpm/commit/3b7e01508a19a2fa6e2ebb1b4685dd4e3cbfd785
refs #291

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -396,10 +396,10 @@ func newItemInsideTarGz(out *tar.Writer, content []byte, header *tar.Header) err
 		header.PAXRecords[fmt.Sprintf("APK-TOOLS.checksum.%s", name)] = fmt.Sprintf("%x", hash.Sum(content))
 	}
 	if err := out.WriteHeader(header); err != nil {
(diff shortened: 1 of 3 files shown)


djgilcrease (approved):
LGTM, just need to fix the mtime test to not compare time.

nxadm:
It looks good to me, thx:

/staging/debs # apk add rakudo-pkg
(1/1) Installing rakudo-pkg (2020.12.0-03)
OK: 141 MiB in 57 packages

The compressed pkg is now 14.9 MB.

caarlos0:
Amazing,  thanks for reporting it btw :D

nxadm:
The speed that you guys/gals fixed this: wow.

PR #297, opened 2021-02-16
Don't override mode/mtime file information if already set

### LINKED ISSUE #296: File info mode and mtime ignored in yaml file
The file information taken from stat syscall at https://github.com/goreleaser/nfpm/blob/2da804eb014f2d44655b58a7ea73276627d3d891/files/files.go#L71-L76
has the effect to always ignore the file_info.mode which could be set in the yaml file, this is also true for mtime

cclerget:
Honor mode and mtime value set in yaml file.

- Fixes #296


djgilcrease (approved):
LGTM

PR #299, opened 2021-02-16
fix: windows globbing

magJ:
Replace windows path separator characters with unix style separators, prior to globbing.
This fixes issues with running goreleaser on windows.


magJ:
I guess windows tests are included in the test coverage stats?

caarlos0:
I think this would fit better on https://github.com/goreleaser/fileglob

can you open a PR there? Also would like to add some tests (we have windows on CI there too)

PR #317, opened 2021-04-11
feat: Support for deb's config maintenance script

joakimkarlsson:
With debconf, the config script is responsible for asking any questions necessary to configure a package [1]. Without this, we're forced to ask questions in postinst, which is against best practices:

> Avoid asking questions in the postinst. Instead, the config script should ask questions using debconf, so that pre-configuration will work.

This adds the possibility to include a config script when using debconf like so:

```yaml
deb:
  scripts:
    templates: ./my/templates
    config: ./my/config
```

[1] https://manpages.debian.org/jessie/debconf-doc/debconf-devel.7.en.html

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -434,6 +434,10 @@ func createControl(instSize int64, md5sums []byte, info *nfpm.Info) (controlTarG
 		fileName: "templates",
 		mode:     0o644,
 	}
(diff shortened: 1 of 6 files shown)


djgilcrease (approved):
LGTM

caarlos0 (approved):
lgtm, can you also add it to the docs?

caarlos0 (approved):
thanks!

PR #320, opened 2021-04-15
fix(ci): speed up rpm tests

caarlos0:
hopefully using fastest mirror and delta rpm it can be faster...


djgilcrease:
did not help with the speed at all, it looks like the `arm64` servers are just SLOW :(

caarlos0:
> did not help with the speed at all, it looks like the `arm64` servers are just SLOW :(

yes :(((

PR #325, opened 2021-04-23
feat: overhaul cmd and release process

caarlos0:
- use cobra so its more like goreleaser
- remove gocenter from goproxy
- using "NFPM is a simple, 0-dependencies, deb, rpm and apk packager." as description in several places
- completions for bash, fish and zsh (also in the brew tap and on the packages)
- cmd docs on website
- cmd code is pretty much ready to accept unit tests (although I haven't added any yet)
- using commit timestamp et al on ldflags so the build is reproducible
- using zip for windows archives
- install the apk package inside the docker image, so we have completions et al there as well (although they wont work as we don't create completions for ash)
- create a discussion for each release

overall, mimic a good deal of config/structure from goreleaser into nfpm.


djgilcrease (approved):
LGTM

PR #353, opened 2021-07-14
fix: deb/apk packager set owner & group failed; feat: add EmptyFolder struct.

Serinalice:
- Bug: During testing, I found that the owner and group settings of content do not take effect on apk and debian.
- Fix: `FileInfo`'s `Sys()` funciton returns `nil`, when use `tar.FileInfoHeader(file, file.Source)` the `Uname` and `Gname` attributes of the header will not be set, so I set them manually.
- Feature: Users should be able to manually control the metadata of the empty folders. For this reason, I added the `EmptyFolder` struct which replaces `string`.

### DIFF
--- a/www/docs/install.md
+++ b/www/docs/install.md
@@ -64,7 +64,7 @@ curl -sfL https://install.goreleaser.com/github.com/goreleaser/nfpm.sh | sh
 go install github.com/goreleaser/nfpm/v2/cmd/nfpm@latest
 ```
 
-### Manually
+### manually
 
 Download the pre-compiled binaries from the [releases page][releases] and copy them to the desired location.
 


caarlos0:
hey, thanks for the PR!

Will review it shortly :)

> * Feature: Users should be able to manually control the metadata of the empty folders. For this reason, I added the `EmptyFolder` struct which replaces `string`.

meanwhile, can we split this into another PR + add related docs?

files/files.go
```diff
@@ -114,6 +114,36 @@ func (c *Content) Sys() interface{} {
...
+	Path  string      `yaml:"path,omitempty"`
+	Owner string      `yaml:"owner,omitempty"`
+	Group string      `yaml:"group"`
+	Mode  os.FileMode `yaml:"mode,omitempty"`
+	MTime time.Time   `yaml:"mtime,omitempty"`
+}
+
+type EmptyFolders []*EmptyFolder
+
+func (f *EmptyFolder) WithFolderInfoDefaults() *EmptyFolder {
```
caarlos0, line 127: this should probably happen on the "global" `WithDefaults` function... iterate the empty folders list and set defaults as needed...

nfpm.go
```diff
@@ -248,18 +248,18 @@ func (i *Info) parseSemver() {
...
-	Deb          Deb            `yaml:"deb,omitempty" jsonschema:"title=deb-specific settings"`
-	APK          APK            `yaml:"apk,omitempty" jsonschema:"title=apk-specific settings"`
+	Replaces     []string           `yaml:"replaces,omitempty" jsonschema:"title=replaces directive,example=nfpm"`
+	Provides     []string           `yaml:"provides,omitempty" jsonschema:"title=provides directive,example=nfpm"`
+	Depends      []string           `yaml:"depends,omitempty" jsonschema:"title=depends directive,example=nfpm"`
+	Recommends   []string           `yaml:"recommends,omitempty" jsonschema:"title=recommends directive,example=nfpm"`
+	Suggests     []string           `yaml:"suggests,omitempty" jsonschema:"title=suggests directive,example=nfpm"`
+	Conflicts    []string           `yaml:"conflicts,omitempty" jsonschema:"title=conflicts directive,example=nfpm"`
+	Contents     files.Contents     `yaml:"contents,omitempty" jsonschema:"title=files to add to the package"`
+	EmptyFolders files.EmptyFolders `yaml:"empty_folders,omitempty" jsonschema:"title=empty folders to be created when installing the package,example=/var/log/nfpm"`
```
caarlos0, line 258: example in jsonschema is no longer valid as this is now a struct.

files/files.go
```diff
@@ -114,6 +114,36 @@ func (c *Content) Sys() interface{} {
 	return nil
 }
 
+type EmptyFolder struct {
+	Path  string      `yaml:"path,omitempty"`
+	Owner string      `yaml:"owner,omitempty"`
+	Group string      `yaml:"group"`
+	Mode  os.FileMode `yaml:"mode,omitempty"`
+	MTime time.Time   `yaml:"mtime,omitempty"`
```
caarlos0, line 122: missing the jsonschema tags

Serinalice: I think maybe it is a good idea to keep the `EmptyFolder` consistent with `Content`/`ContentInfo`. 
Are jsonschema tags necessary?

nfpm.go
Serinalice: Yeah,I will fix this.

files/files.go
Serinalice: Oh,Sorry...I missed the function `WithDefaults `. I will reconsider that and update.

Serinalice:
> hey, thanks for the PR!
> 
> Will review it shortly :)
> 
> > * Feature: Users should be able to manually control the metadata of the empty folders. For this reason, I added the `EmptyFolder` struct which replaces `string`.
> 
> meanwhile, can we split this into another PR + add related docs?

Yes, no problem.

files/files.go
djgilcrease: I would try to make this Just part of the content structure with an `emptyfolder` type. And leave the current empty folder structure alone for backwards compatibility. We can then depreciate the empty folder structure.

caarlos0: we could also write a custom decoder for this type, so the old format (string only) still works...

PR #359, opened 2021-08-06
fix: deb and apk packager set file's mode failed

Serinalice:
Because `Sys()` of `Content` returns nil, so `tar.FileInfoHeader(file, file.Source)` can't populate additional fields to set `Uname` and `Gname`.
```
if sys, ok := fi.Sys().(*Header); ok {
		// This FileInfo came from a Header (not the OS). Use the
		// original Header to populate all remaining fields.
		h.Uid = sys.Uid
		h.Gid = sys.Gid
		h.Uname = sys.Uname
		h.Gname = sys.Gname
		h.AccessTime = sys.AccessTime
		h.ChangeTime = sys.ChangeTime
		if sys.Xattrs != nil {
			h.Xattrs = make(map[string]string)
			for k, v := range sys.Xattrs {
				h.Xattrs[k] = v
			}
		}
		if sys.Typeflag == TypeLink {
			// hard link
			h.Typeflag = TypeLink
			h.Size = 0
			h.Linkname = sys.Linkname
		}
		if sys.PAXRecords != nil {
			h.PAXRecords = make(map[string]string)
			for k, v := range sys.PAXRecords {
				h.PAXRecords[k] = v
			}
		}
	}
```

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -472,6 +472,8 @@ func copyToTarAndDigest(file *files.Content, tw *tar.Writer, sizep *int64) error
 	}
 
 	header.Name = files.ToNixPath(file.Destination[1:])
+	header.Uname = file.FileInfo.Owner
+	header.Gname = file.FileInfo.Group
 	if err = newItemInsideTarGz(tw, contents, header); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


caarlos0:
can you add a test for this?

djgilcrease (approved):
LGTM

djgilcrease:
Once you squash the commits I will merge this in.

caarlos0:
> Once you squash the commits I will merge this in.

I would say we should add tests for this as well, so we wont break it again in the future

PR #363, opened 2021-09-02
feat: compression option with xz support for deb

### LINKED ISSUE #361: XZ-compressed Debian packages
I was reading over the Debian repository format, which lead to me looking into their current package formats and discovered that they now seem to require XZ support. Inspecting random packages in official Debian repos, I found the .deb  and the data.tar inside are both XZ-compressed, so it looks like they want this to be the default now.

There's already some support for creating .tar.xz archives in GoReleaser, so it's not a huge leap to also apply it to Debian packages. Making it default is probably not going to be problematic, but having the option to support old clients which only read Gzip is of course useful.

erikgeiser:
This PR adds a `compression` option for `deb` files. Currently the following values are supported:

* `gzip` (default)
* `xz` (future default)
* `none` (no compression)

I think the best way to move forward is to keep `gzip` as default for now and wait for feedback from the users that use `xz`. If everything works great (and if `xz` build times are acceptable), we should make `xz` the default in a following release.

Closes #361.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -91,7 +91,7 @@ func (*Deb) Package(info *nfpm.Info, deb io.Writer) (err error) { // nolint: fun
 		return err
 	}
 
-	dataTarGz, md5sums, instSize, dataTarballName, err := createDataTarball(info)
+	dataTarball, md5sums, instSize, dataTarballName, err := createDataTarball(info)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


deb/deb.go
```diff
@@ -90,7 +91,7 @@ func (*Deb) Package(info *nfpm.Info, deb io.Writer) (err error) { // nolint: fun
 		return err
 	}
 
-	dataTarGz, md5sums, instSize, err := createDataTarGz(info)
+	dataTarGz, md5sums, instSize, dataTarballName, err := createDataTarball(info)
```
caarlos0, line 94: `dataTarGz` could be named `dataTarball` now I think

caarlos0 (approved):
minor nitpick, lgtm otherwise

great work, thanks!

deb/deb.go
erikgeiser: Thanks, I found a lot of other places too where I did not update the naming. It should all be consistent now.

caarlos0 (approved):

erikgeiser:
I've prepared another PR that changes the default compression algorithm and the corresponding documentation. I will publish it as soon as this PR is merged, so you can easily update the default in another release if everything goes well.

Issue #100, opened 2019-11-05
validate required fields


caarlos0:
maybe we should add a step to validate that required fields were provided?

I'm not sure about debs, but RPMs seem to like to complain about missing required fields :) 

see https://github.com/goreleaser/goreleaser/issues/1229 for more context

erikgeiser:
We should also print warnings to `stderr` when top level config options are not supported by the package format that is currently being built. If a configuration option is top level and not package specific, users would otherwise expect it to have an effect on all formats. I think `apk` files currently don't support some options.

caarlos0:
this was fixed in #372

PR #372, opened 2021-10-08
feat: enforce packager defaults

drewstinnett:
Not completely addressing #100, but giving it a first go!

This PR adds in some `deb` specific defaults, specifically around:
* Although the docs are a little murky ([some](https://www.debian.org/doc/debian-policy/ch-archive.html#priorities) say this field must be set, [others](https://www.debian.org/doc/debian-policy/ch-controlfields.html#binary-package-control-files-debian-control) say it is just recommended), there are some 3rd party tools that will barf if priority isn't set, so it feels like a good idea to set this to a safe default on all debs
* Maintainer is a 'mandatory' field...leaving it unset still results in a usable deb package, however installing the deb package prints a nasty warning about not having a maintainer

Would love any feedback/suggestions around making this better.

One thing I'm a little torn on is setting the `Maintainer` field to a bogus value. Even though this field is marked as mandatory in some of the deb documentation, but I would hate to break existing setups by making nfpm for real require it, instead of just setting a bogus value.

When (or if 🤣) you're cool with the implementation details, I'll be happy to extend this to rpm/apk too. Thanks!!

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -167,7 +167,7 @@ func (*Deb) SetPackagerDefaults(info *nfpm.Info) {
 	// if in the long run we should be more strict about this and error when
 	// not set?
 	if info.Maintainer == "" {
-		log.Println("DEPRECATION WARNING: Unset 'maintainer' field on deb packages is deprecated and will be removed in a future version")
+		log.Println("DEPRECATION WARNING: Leaving the 'maintainer' field unset will not be allowed in a future version")
 		info.Maintainer = "Unset Maintainer <unset@localhost>"
 	}
 }


erikgeiser:
My preferred way would be this: Set a bogus value but print out a "deprecation" warning. Don't document the bogus value but document that leaving is empty is "deprecated". After a while (or with `v3` if necessary) we remove the bogus value and force the maintainer to be set.

Maybe another word would fit this purpose better than "deprecation", though. 

What do you you think?

drewstinnett:
@erikgeiser , that would work from my end! I will update the PR

erikgeiser:
A safe value for the priority would be `optional` I guess. According to the [docs](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities): *"This is the default priority for the majority of the archive. Unless a package should be installed by default on standard Debian systems, it should have a priority of optional. Packages with a priority of optional may conflict with each other."*

Oh by the way, maybe the bogus value should be `Unset Maintainer <unset@localhost>`?

PR #388, opened 2021-10-23
Fix: Improve performance of content destination collision detection.

### LINKED ISSUE #387: Improve performance of content collision detection
This issue was raised by @mjmjelde in https://github.com/goreleaser/nfpm/discussions/385.

In `ExpandContentGlobs` we try building the content list by appending only elements with unique destinations. This increases the complexity in way that it does not scale for large amount of files. I hacked together a POC solution where the collision detection is performed once near the end before sorting. The benchmarks are promising:
```
========= 10000 Content Entries =========
goos: darwin
goarch: amd64
pkg: github.com/goreleaser/nfpm/v2/files
cpu: Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
BenchmarkExpandContentGlobs-4       	       1	8943350742 ns/op	5948665080 B/op	 2963056 allocs/op
BenchmarkExpandContentGlobsFast-4   	       2	 690795220 ns/op	503257972 B/op	 1260368 allocs/op
PASS
ok  	github.com/goreleaser/nfpm/v2/files	11.305s

========= 100000 Content Entries =========
goos: darwin
goarch: amd64
pkg: github.com/goreleaser/nfpm/v2/files
cpu: Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
BenchmarkExpandContentGlobs-4       	       1	1116330817406 ns/op	545598423120 B/op	185560019 allocs/op
BenchmarkExpandContentGlobsFast-4   	       1	25918381508 ns/op	41183230936 B/op	12604194 allocs/op
PASS
ok  	github.com/goreleaser/nfpm/v2/files	1142.497s
```
For large packages, this can be more than a 40x speedup.

I'll submit a PR if I find the time to clean up the very hacky POC.

erikgeiser:
This PR fixes #387 and #393 by reducing the algorithmic complexity of the content destination collision detection. Previously when appending an element, the destination was compared to all other elements in the content list. Now we simply append and check for collisions once at the end with a single pass through all elements.

For 150000 files with random destinations, the performance compares as follows:
```
Before: BenchmarkExpandContentGlobs-4   	       1	2982879539110 ns/op
Now   : BenchmarkExpandContentGlobs-4   	       1	9052880061 ns/op
```
**That's a 320x improvement from 50 minutes to 9 seconds for extreme use cases!**

Here's the benchmark code:
```Go
func BenchmarkExpandContentGlobs(b *testing.B) {
	contents := testContents(150000)
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		_, err := files.ExpandContentGlobs(contents, true)
		if err != nil {
			b.Fatalf("expand content globs: %v", err)
		}
	}
}

func testContents(n int) files.Contents {
	contents := make(files.Contents, n)

	for i := 0; i < n; i++ {
		contents[i] = &files.Content{
			Source:      "../testdata/whatever.conf",
			Destination: randomString(),
		}
	}

	return contents
}

const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"

func randomString() string {
	b := make([]byte, 5+rand.Intn(20))
	for i := range b {
		b[i] = letters[rand.Intn(len(letters))]
	}
	return string(b)
}
```

Thank you @mjmjelde for reporting this.

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -116,33 +116,35 @@ func (c *Content) Sys() interface{} {
 
 // ExpandContentGlobs gathers all of the real files to be copied into the package.
 func ExpandContentGlobs(contents Contents, disableGlobbing bool) (files Contents, err error) {
+	options := []fileglob.OptFunc{fileglob.MatchDirectoryIncludesContents}
+	if disableGlobbing {
+		options = append(options, fileglob.QuoteMeta)
+	}
+
 	for _, f := range contents {
 		var globbed map[string]string
 
 		switch f.Type {
 		case "ghost", "symlink":
 			// Ghost and symlink files need to be in the list, but dont glob them because they do not really exist
-			files, err = appendWithUniqueDestination(files, f.WithFileInfoDefaults())
+			files = append(files, f.WithFileInfoDefaults())
+		default:
+			globbed, err = glob.Glob(f.Source, f.Destination, options...)
 			if err != nil {
 				return nil, err
 			}
 
-			continue
+			files, err = appendGlobbedFiles(files, globbed, f)
+			if err != nil {
+				return nil, err
+			}
 		}
 
-		options := []fileglob.OptFunc{fileglob.MatchDirectoryIncludesContents}
-		if disableGlobbing {
-			options = append(options, fileglob.QuoteMeta)
-		}
-		globbed, err = glob.Glob(f.Source, f.Destination, options...)
-		if err != nil {
-			return nil, err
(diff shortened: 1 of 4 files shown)


erikgeiser:
I added a commit that removes the duplicate `Validate` call to fix #393.

caarlos0:
amazing! thanks!

PR #390, opened 2021-11-03
feat: Directory content type with custom attributes

erikgeiser:
This PR introduces the content type `dir` which can be used to add empty directories or directories with custom attributes such as owner, group or permissions. It also makes `EmptyFolders` obsolete.

Another important aspect is that it can be used to specified directories "owned" by the package when packaging RPMs as only those directories are removed when the package is uninstalled. This use case is also now well documented.

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -404,39 +404,58 @@ func createBuilderData(info *nfpm.Info, sizep *int64) func(tw *tar.Writer) error
 }
 
 func createFilesInsideTarGz(info *nfpm.Info, tw *tar.Writer, created map[string]bool, sizep *int64) (err error) {
+	// create explicit directories first
 	for _, file := range info.Contents {
+		// at this point, we don't are about other types yet
+		if file.Type != "dir" {
+			continue
+		}
+
+		// only consider contents for this packager
 		if file.Packager != "" && file.Packager != packagerName {
 			continue
 		}
+
+		err = tw.WriteHeader(&tar.Header{
+			Name:     file.Destination,
+			Mode:     int64(file.FileInfo.Mode),
+			Typeflag: tar.TypeDir,
+			Format:   tar.FormatGNU,
+			Uname:    file.FileInfo.Owner,
+			Gname:    file.FileInfo.Group,
+			ModTime:  file.FileInfo.MTime,
+		})
+		if err != nil {
+			return err
+		}
+
+		created[strings.TrimPrefix(file.Destination, "/")] = true
+	}
+
+	for _, file := range info.Contents {
+		// only consider contents for this packager
+		if file.Packager != "" && file.Packager != packagerName {
+			continue
+		}
+
+		// create implicit directory structure below the current content
 		if err = createTree(tw, file.Destination, created); err != nil {
 			return err
 		}
+
 		switch file.Type {
 		case "ghost":
 			// skip ghost files in apk
 			continue
+		case "dir":
+			// already handled above
+			continue
 		case "symlink":
 			err = createSymlinkInsideTarGz(file, tw)
-		case "dir":
-			// this .nope is actually not created, because createTree ignore the
-			// last part of the path, assuming it is a file.
-			// TODO: should probably refactor this
-			err = createTree(tw, filepath.Join(file.Destination, ".nope"), created)
-		case "doc":
+		case "doc", "licence", "license", "readme", "config", "config|noreplace":
 			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
-			fallthrough
-		case "licence", "license":
-			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
-			fallthrough
-		case "readme":
-			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
-			fallthrough
-		case "config", "config|noreplace":
-			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
+			// ignoring `emptyFallthrough: remove empty case containing only
+			// fallthrough to default case`
 			fallthrough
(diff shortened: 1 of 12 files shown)


apk/apk.go
```diff
@@ -399,54 +399,70 @@ func createBuilderData(info *nfpm.Info, sizep *int64) func(tw *tar.Writer) error
...
+			// fallthrough to default case`
 			fallthrough
 		default:
 			err = copyToTarAndDigest(file, tw, sizep)
 		}
 		if err != nil {
 			return err
 		}
-		created[file.Source] = true
-		created[file.Destination[1:]] = true
```
erikgeiser, line 449: I don't know why these two lines where there to begin with. The source only matters for symlinks but created is only to avoid creating duplicate implicit directories anyway, right?

apk/apk.go
```diff
@@ -528,7 +532,7 @@ func pathsToCreate(dst string) []string {
 	}
 	// we don't really need to create those things in order apparently, but,
 	// it looks really weird if we don't.
-	var result []string
```
erikgeiser, line 531: This now matches the implementation in `deb.go`.

www/docs/configuration.md
```diff
@@ -53,8 +53,7 @@ priority: extra
 
 # Maintainer.
 # Defaults to empty on rpm and apk
-# Leaving this field empty on 'deb' packages is deprecated, and will be removed
-# in a future release
```
erikgeiser, line 57: This is just a drive-by fix to make it more in line with the deprecation warning.

deb/deb.go
```diff
@@ -166,7 +165,7 @@ func (*Deb) SetPackagerDefaults(info *nfpm.Info) {
 	// if in the long run we should be more strict about this and error when
 	// not set?
 	if info.Maintainer == "" {
-		log.Println("DEPRECATION WARNING: Leaving the 'maintainer' field unset will not be allowed in a future version")
```
erikgeiser, line 169: We don't use `log`, elsewhere, so I changed it to print to `os.Stderr` just like the `empty_folders` deprecation warning.

djgilcrease (approved):
LGTM

apk/apk.go
```diff
@@ -396,54 +396,70 @@ func createBuilderData(info *nfpm.Info, sizep *int64) func(tw *tar.Writer) error
...
-
-		// handle Files
 		return createFilesInsideTarGz(info, tw, created, sizep)
 	}
 }
 
 func createFilesInsideTarGz(info *nfpm.Info, tw *tar.Writer, created map[string]bool, sizep *int64) (err error) {
+	// create explicit directories first
+	for _, file := range info.Contents {
+		// at this point, we don't are about other types yet
```
caarlos0, line 406:
```suggestion
		// at this point, we don't care about other types yet
```

apk/apk.go
```diff
@@ -396,54 +396,70 @@ func createBuilderData(info *nfpm.Info, sizep *int64) func(tw *tar.Writer) error
...
-		case "readme":
+		case "doc", "licence", "license", "readme", "config", "config|noreplace":
 			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
-			fallthrough
-		case "config", "config|noreplace":
-			// nolint:gocritic
-			// ignoring `emptyFallthrough: remove empty case containing only fallthrough to default case`
+			// ignoring `emptyFallthrough: remove empty case containing only
+			// fallthrough to default case`
```
caarlos0, line 455: can we remove all this and let it all go to the `default`?

apk/apk.go
caarlos0: I don't know either 🤔  but I think if tests are passing its probably something we didn't need indeed..

apk/apk_test.go
```diff
@@ -480,3 +485,126 @@ func TestPackageSymlinks(t *testing.T) {
...
+	}
+
+	require.NoError(t, info.Validate())
+
+	var buf bytes.Buffer
+	size := int64(0)
+	err := createFilesInsideTarGz(info, tar.NewWriter(&buf), make(map[string]bool), &size)
+	require.NoError(t, err)
+
+	// for debs all implicit or explicit directories are created in the tarball
```
caarlos0, line 521: should it say for `apks` instead of debs?

deb/deb.go
caarlos0: thinking out loud here: maybe we could allow the caller to pass a `io.Writer` or some other interface as to "where should we print the warnings" 🤔

deb/deb.go
```diff
@@ -271,10 +267,41 @@ func createSymlinkInsideTar(file *files.Content, out *tar.Writer) error {
 
 func createFilesInsideDataTar(info *nfpm.Info, tw *tar.Writer,
 	created map[string]bool) (md5buf bytes.Buffer, instSize int64, err error) {
+	// create explicit directories first
 	for _, file := range info.Contents {
+		// at this point, we don't are about other types yet
```
caarlos0, line 272:
```suggestion
		// at this point, we don't care about other types yet
```

files/files.go
```diff
@@ -125,7 +152,7 @@ func ExpandContentGlobs(contents Contents, disableGlobbing bool) (files Contents
 		var globbed map[string]string
 
 		switch f.Type {
-		case "ghost", "symlink":
+		case "ghost", "symlink", "dir":
 			// Ghost and symlink files need to be in the list, but dont glob them because they do not really exist
```
caarlos0, line 156: comment is now outdated since we added `dir` as well

caarlos0 (approved):
just some nitpicks, but looks good overall!

awesome work!

deb/deb.go
erikgeiser: Another idea, a package level `io.Writer` for warnings that defaults to `os.Stderr` that can be overwritten globally. It's simple enough to add it to this PR.

caarlos0:
awesome work as always, thanks @erikgeiser !

PR #391, opened 2021-11-04
fix docker exec example

hellt:
Hi

1. looks like the $PWD should be bind mounted as `/tmp`, otherwise the `/tmp` won't be visible on the host system
2. I set the container workdir to `tmp` so that the relative paths in foo.yml can be resolved


caarlos0:
Thanks!

PR #395, opened 2021-11-05
Fix: Allow complex file modes for debs and apks.

### LINKED ISSUE #392: File Mode issue when using setuid bits
Howdy,

I have created an application that requires the setuid bit to be on one of the executable.  However, when I have a custom file_info mode set with the setuid bit set, it does not actually get properly set in the deb or rpm package that is created.  In fact, some weird permissions appear for the file in the deb package and the rpm package is not even able to install.

This is what I have in my nfpm.yaml file:
```
contents:
  - src: dist/myexec
    dst: /usr/local/bin/myexec
    file_info:
      mode: 4555
      owner: root
      group: root
```

But this is what permission actually show up in the package:
```
root@0db5c89952b8:/# ls -la /usr/local/bin/myexec
-rwx--x-wx. 1 root root 50326500 Nov  4 16:01 /usr/local/bin/myexec
```

erikgeiser:
This PRs fixes #392 by passing the full file mode instead of just the permissions (which is the mode masked with `0o777`) to the `tar` header. It also added acceptance tests with suid bit for all packagers and documentation that reminds users to specify the mode in octal.

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -471,6 +471,9 @@ func copyToTarAndDigest(file *files.Content, tw *tar.Writer, sizep *int64) error
 		return err
 	}
 
+	// tar.FileInfoHeader only uses file.Mode().Perm() which masks the mode with
+	// 0o777 which we don't want because we want to be able to set the suid bit.
+	header.Mode = int64(file.Mode())
 	header.Name = files.ToNixPath(file.Destination[1:])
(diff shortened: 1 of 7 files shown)


djgilcrease (approved):
LGTM

caarlos0:
noice, thanks!

PR #396, opened 2021-11-05
Feature: Allow platform specific arch overrides.

### LINKED ISSUE #367: [deb] "arm" mistakenly converted to armhf
It seems that nfpm assumes `arm` means `armhf` [see here](https://github.com/goreleaser/nfpm/blob/master/deb/deb.go#L39). 

I don't see a way to override this behavior.

I'd be happy to send a pull request but I'm not sure what the best strategy should be to allow users to customize the behavior. Thoughts?

Per the [Debian Policy Documentation](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#s-arch-spec) the `dpkg-architecture` command will list valid architectures

Here is the output from my debian machine:
```
$ dpkg-architecture -L | grep arm
<snip>
armhf
armel
arm64ilp32
armeb
arm
arm64
<snip>
```

erikgeiser:
This PR fixes #367 by allowing platform specific arch overrides. The documentation now also clarifies that the `arch` field is translated and references the override options.

I also removed `arch` defaulting to `amd64` for APKs as it was not consistent with the other packagers and it should be specified in any case in order to avoid invalid packages (`amd64` packages without `amd64` specific contents or with an `arm` binary).

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -71,6 +71,19 @@ var archToAlpine = map[string]string{
 	// "s390x":  "???",
 }
 
+func ensureValidArch(info *nfpm.Info) *nfpm.Info {
+	if info.APK.Arch != "" {
+		info.Arch = info.APK.Arch
+	} else {
+		arch, ok := archToAlpine[info.Arch]
+		if ok {
+			info.Arch = arch
+		}
+	}
+
+	return info
+}
+
 // Default apk packager.
 // nolint: gochecknoglobals
 var Default = &Apk{}
@@ -79,12 +92,7 @@ var Default = &Apk{}
 type Apk struct{}
 
 func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
-	// TODO: verify this
-	arch, ok := archToAlpine[info.Arch]
-	if !ok {
-		arch = info.Arch
-	}
-
+	info = ensureValidArch(info)
 	version := info.Version
 	if info.Release != "" {
 		version += "-" + info.Release
@@ -98,20 +106,12 @@ func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
 		version += "+" + info.VersionMetadata
 	}
 
-	return fmt.Sprintf("%s_%s_%s.apk", info.Name, version, arch)
+	return fmt.Sprintf("%s_%s_%s.apk", info.Name, version, info.Arch)
 }
 
 // Package writes a new apk package to the given writer using the given info.
 func (*Apk) Package(info *nfpm.Info, apk io.Writer) (err error) {
-	if info.APK.APKArch != "" {
-		info.Arch = info.APK.APKArch
-	} else {
-		arch, ok := archToAlpine[info.Arch]
-		if ok {
-			info.Arch = arch
-		}
-	}
-
+	info = ensureValidArch(info)
 	if err = info.Validate(); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


apk/apk.go
```diff
@@ -103,13 +103,15 @@ func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
...
-	if info.Arch == "" {
-		info.Arch = archToAlpine["amd64"]
+	if info.APK.APKArch != "" {
+		info.Arch = info.APK.APKArch
+	} else {
+		arch, ok := archToAlpine[info.Arch]
+		if ok {
+			info.Arch = arch
+		}
 	}
```
djgilcrease, line 113: Looks like lines 83-86 might need to be changed / cleaned up as well?
Same for the deb.go as well.

Maybe add a `func ensureValidArch` like exists in rpm.go for apk and deb?

rpm/rpm.go
```diff
@@ -61,10 +61,15 @@ var archToRPM = map[string]string{
 }
 
 func ensureValidArch(info *nfpm.Info) *nfpm.Info {
-	arch, ok := archToRPM[info.Arch]
-	if ok {
-		info.Arch = arch
+	if info.APK.APKArch != "" {
+		info.Arch = info.APK.APKArch
```
djgilcrease, line 65: Guessing this should be `info.RPM.Arch`?

nfpm.go
```diff
@@ -263,6 +263,7 @@ type Overridables struct {
 
 // RPM is custom configs that are only available on RPM packages.
 type RPM struct {
+	RPMArch     string       `yaml:"rpm_arch,omitempty" jsonschema:"title=architecture in rpm nomenclature"`
```
djgilcrease, line 266:
```go
Arch     string       `yaml:"arch,omitempty" jsonschema:"title=architecture in rpm nomenclature"`
```

nfpm.go
```diff
@@ -288,6 +289,7 @@ type RPMSignature struct {
 }
 
 type APK struct {
+	APKArch   string       `yaml:"apk_arch,omitempty" jsonschema:"title=architecture in apk nomenclature"`
```
djgilcrease, line 292:
```go
Arch     string       `yaml:"arch,omitempty" jsonschema:"title=architecture in apk nomenclature"`
```

nfpm.go
```diff
@@ -305,6 +307,7 @@ type APKScripts struct {
 
 // Deb is custom configs that are only available on deb packages.
 type Deb struct {
+	DebArch     string       `yaml:"deb_arch,omitempty" jsonschema:"title=architecture in deb nomenclature"`
```
djgilcrease, line 310:
```go
Arch     string       `yaml:"arch,omitempty" jsonschema:"title=architecture in deb nomenclature"`
```

djgilcrease (changes requested):
Over all looks good, just a few thing

nfpm.go
erikgeiser: Oh yeah, I thought `Arch` was in `Overridables` but it actually isn't.

apk/apk.go
erikgeiser: Done

rpm/rpm.go
erikgeiser: Correct

erikgeiser:
> Over all looks good, just a few thing

Thanks, good catches, I kinda botched this PR a bit

caarlos0:
Noice, thanks!

PR #398, opened 2021-11-07
fix: added more arches

caarlos0:
added more arches, improved test coverage a bit

### DIFF
--- a/deb/deb_test.go
+++ b/deb/deb_test.go
@@ -1086,10 +1086,9 @@ func TestEmptyButRequiredDebFields(t *testing.T) {
 }
 
 func TestArches(t *testing.T) {
-	info := exampleInfo()
-
 	for k := range archToDebian {
 		t.Run(k, func(t *testing.T) {
+			info := exampleInfo()
 			info.Arch = k
 			info = ensureValidArch(info)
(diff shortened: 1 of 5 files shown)


deb/deb_test.go
```diff
@@ -1084,3 +1084,21 @@ func TestEmptyButRequiredDebFields(t *testing.T) {
...
+	for k := range archToDebian {
+		t.Run(k, func(t *testing.T) {
+			info.Arch = k
+			info = ensureValidArch(info)
+			require.Equal(t, archToDebian[k], info.Arch)
+		})
+	}
+
+	t.Run("override", func(t *testing.T) {
+		info.Deb.Arch = "foo64"
```
erikgeiser, line 1100: I'd add `info.Arch = "amd64"` or something like this before this line. Otherwise this test depends on whatever value `info.Arch` was set to in the previous subtests. This also makes it more clear in what way the override behavior is tested.

erikgeiser (approved):
LGTM, just one minor nitpick

deb/deb_test.go
caarlos0: fixed it to use a new exampleInfo on every subtest :)

PR #399, opened 2021-11-12
feat: add deprecation notice interface

caarlos0:
as discussed in another issue

this would allow, for example, goreleaser, to set its own deprecation notice "provider" to nfpm...


erikgeiser:
I intuition tells me to simplify it. What's the advantage compared to this?
```go
var DeprecationNoticeWriter io.Writer = os.Stderr

 // Notice will notice a deprecation with the previously set noticer.
func Notice(s string) {
     DeprecationNoticeWriter.Write(s)
 }

func Noticef(format string, a ...interface{}) {
    DeprecationNoticeWriter.Write(fmt.Sprintf(format, a...)
}
```

caarlos0:
on my impl it would be easier to use, say, a logger library... but its fair, the caller can always implement a simple writer that logs...

will change it :)

PR #405, opened 2021-11-18
feat: add expand-env support for more fields

novasharper:
also expand environment variables in the package name, and src/dst
for package contents entries

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -65,6 +65,13 @@ func (c Contents) ContainsDestination(dst string) bool {
 	return false
 }
 
+func (c Contents) ExpandEnvVars(envMappingFunc func(string) string) {
+	for _, content := range c {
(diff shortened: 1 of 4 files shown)


djgilcrease:
So the issue with doing this on file contents is `$` is a file filename or path character, and the default behavior is to replace all potential env vars with an empty string if not set, so we would need a custom env mapping function that does no replacement if the variable was not set. I am disinclined to approve with a custom function as then we would need to support it for all potential uses and forms of env vars.

Maybe update the `Contents` struct to have an expand_env bool to determine if we should expand or not? @caarlos0 @erikgeiser  what do you think?

caarlos0:
yeah, I agree with @djgilcrease , this is likely a no-go.

thanks for the PR and the effort though. 🙏

novasharper:
@caarlos0 So not even with adding an `expand_env` field to the `Content` struct?

PR #410, opened 2021-12-01
fix: Fix destinations for files containing matchers when globbing is disabled.

rtpt-erikgeiser:
This PR fixes a bug where file sources containing matchers result in the files being created inside folders of the same name as the file when globbing is disabled.

**Example:**
```
src: foo/{bar}/config.json
dst: /etc/config.json
```
resulted in this:
```
src: foo/{bar}/config.json
dst: /etc/config.json/config.json
```
when it should have actually just remain unchanged.

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/goreleaser/fileglob"
 	"github.com/goreleaser/nfpm/v2/internal/glob"
 )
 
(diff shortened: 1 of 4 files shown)


rtpt-erikgeiser:
I have no idea why, but the Windows CI task fails due to missing `gcc`. I don't think that's related to this RP.

djgilcrease (approved):
Lgtm

caarlos0:
nice, thanks!

PR #411, opened 2021-12-01
fix: Prevent duplicate folders from being created.

rtpt-erikgeiser:
Currently a package can contain duplicate folders depending on the way a user specified a dir-type content destination (folder name). This PR fixes that by keeping track of created folders in the already existing `created` map with a **consistent naming convention**. It also adds related tests.

Another bug that's also fixed by this PR is that auto-created folders did not have a owner username and group name.


djgilcrease (approved):
Lgtm

caarlos0:
good catch, thanks!

Issue #423, opened 2021-12-10
Missing test `epoch`


4censord:
We seem to be missing a testcase where the `epoch` field is set.

erikgeiser:
This is tested in `TestRPM`, `TestDebEpoch` but not in for the `apk` packager. Honestly. the `apk` packager is missing a lot of tests to achieve testing parity with `deb` and `rpm`.

4censord:
Oh, i see. I had only looked at the acceptance tests.

erikgeiser:
So we'd need an epoch unit test for the `apk` packager and we should extend `core.complex.yaml` to include the epoch and test for it in the Dockerfiles.

4censord:
should that be two seperate pull requests or one combined one?

erikgeiser:
That can both be done in a single PR.

caarlos0:
closing as stale

PR #424, opened 2021-12-11
Test: more generic tests.

4censord:
Simple things:
* a empty folder
* Epoch is set
* a broken symlink
* a path longer than 256 characters
* A path with windows style seperators
* A path starting with a windows style `C:\`

File Validation tests
* a file that is included twice
* a file with a name longer than 256 characters
* a file with a space
* a file with a leading space
* a file with a newline character
* a file with null character

Path validation tests
* A src path with `..`
* A dst path with `../..`, thus escaping above /
* 5 different paths that are valid, but not sanitised
  * with `..`
  * multiple `/`
  * `/./`

They are not yet tested as part of the automated acceptance tests.

I think we need to upgrade the acceptance-test.go to allow for tests that are expected to fail.

- [ ] Allow `expect fail` tests in acceptance-test.go
- [ ] Add all tests to acceptance tests

### DIFF
--- a/testdata/acceptance/core.epoch.yaml
+++ b/testdata/acceptance/core.epoch.yaml
@@ -1,0 +1,10 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+epoch: 2
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: /usr/local/bin/fake
--- a/testdata/acceptance/core.file-twice.yaml
+++ b/testdata/acceptance/core.file-twice.yaml
@@ -11,6 +11,5 @@ release: "symlink"
 contents:
   - src: ./testdata/whatever.conf
     dst: /etc/foo/whatever.conf
-  - src: /etc/foo/whatever.conf
-    dst: /path/to/symlink
-    type: symlink
+  - src: ./testdata/whatever.conf
+    dst: /etc/foo/whatever.conf
--- a/testdata/acceptance/core.filename-longer-256.yaml
+++ b/testdata/acceptance/core.filename-longer-256.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: /usr/verryyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyloooooooooooooooooooooooooooooooooooooooooooooooooooooooongfilename123456780000000000000000000000000000000000000000000000000000000000
--- a/testdata/acceptance/core.filename-with-leading-whitespace.yaml
+++ b/testdata/acceptance/core.filename-with-leading-whitespace.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: "/usr/ filename"
--- a/testdata/acceptance/core.filename-with-newline.yaml
+++ b/testdata/acceptance/core.filename-with-newline.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: "/usr/file\nname"
--- a/testdata/acceptance/core.filename-with-null.yaml
+++ b/testdata/acceptance/core.filename-with-null.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: "/usr/file\0name"
--- a/testdata/acceptance/core.filename-with-whitespace.yaml
+++ b/testdata/acceptance/core.filename-with-whitespace.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
+    dst: "/usr/file name"
--- a/testdata/acceptance/core.folder.yaml
+++ b/testdata/acceptance/core.folder.yaml
@@ -7,10 +7,7 @@ maintainer: "Foo Bar"
 vendor: "foobar"
 homepage: "https://foobar.org"
 license: "MIT"
-release: "symlink"
+release: "folder"
 contents:
-  - src: ./testdata/whatever.conf
-    dst: /etc/foo/whatever.conf
-  - src: /etc/foo/whatever.conf
-    dst: /path/to/symlink
-    type: symlink
+  - dst: /path/to/folder
+    type: dir
--- a/testdata/acceptance/core.long-path-256.yaml
+++ b/testdata/acceptance/core.long-path-256.yaml
@@ -1,0 +1,9 @@
+---
+name: foo
+arch: "${BUILD_ARCH}"
+version: 1.2.3
+license: MIT
+maintainer: "Foo Bar"
+contents:
+  - src: ./testdata/fake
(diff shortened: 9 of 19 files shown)


4censord:
I don't know how to add those to the automated acceptance tests.
I would appreciate some guidance/help on how those work.

4censord:
This would close #423

erikgeiser:
Hey, thanks for your contribution. It is always a great idea to think about what we can or should test.

However, I'm not sure what the benefits of intentionally failing acceptance tests are. If broken packages can be produced we should already prevent that with unit tests. Acceptance tests should rather test if the features we support do actually work on a real system.

I've also gathered some thoughts about the test cases:


Simple things:

* **a empty folder:** this one is already covered by the unit tests which check if the entry for a directory is created with the correct type mode and if it can be extracted again from the `tar`. Testing this again in an acceptance test would be more of a test of `dpkg`/`apt`  than of `nfpm`.
* **Epoch is set:** This mechanisms works identically to, for example, the version field which is already tested. There is some value to an acceptance test (checking if we spelled the field name correctly such that `dpkg` recognizes it) but acceptance tests are more expensive so including it in a carefully constructed unit test would maybe be better.
* **a broken symlink:** This is the responsibility of the user. Afaik `nfpm` allows for valid and invalid symlinks and we shouldn't change that because we don't know what the user wants to achieve with it. For example, an `rpm` could be built with a symlink to valid files in Fedora that just don't exist in CentOS.
* **a path longer than 256 characters:** If this violates `cpio`/`tar` spec `nfpm` should error out, then there is nothing to test. If some filesystems don't accept this, this is also in the responsibility of the user because users maybe build for distros with alternative file systems.
* **A path with windows style seperators:** This should also be validated in a unit test. I don't know what actually happens. If a file containing `\` in the name is created we could warn the user but that also might be a false positive.
* **a file with a leading space/A path with windows style seperators:** Files are binary blobs and can contain anything. This would just test the Go standard library and `dpkg`/`rpm`.
* **A path starting with a windows style `C:/`:** We could handle that with an custom error but even then, this should be a unit test.

File Validation tests

* **a file that is included twice:** Already tested in `TestCollision` in `files_test.go`.
* **a file with a name longer than 256 characters:** See above
* **file with a space/leading space/newline/null** See above
* **5 different paths that are valid:**  I think it is reasonable to just assume that the Go standard library (`io`/ `tar`, ...) handles all this fine.

Path validation tests

* **A src path with:** Already tested in multiple tests.
* **5 different paths that are valid, but not sanitized:** We could add some sanity checks in `files.go` and corresponding unit tests in `files_test.go`.
* **A dst path with `../..`, thus escaping above `/`:** There is no escaping the filesystem root afaik, but we already clean the destination anyway but we could still add a unit test in `files_test.go`.

* **A src path with:** Already tested in multiple tests.
* **5 different paths that are valid, but not sanitized:** We could add some sanity checks in `files.go` and corresponding unit tests in `files_test.go`.
* **A dst path with `../..`, thus escaping above `/`:** There is no escaping the filesystem root afaik, but we already clean the destination anyway but we could still add a unit test in `files_test.go`.

I'm happy to discuss any of these points, though.

4censord:
I added those tests because I started building a packager for
Pacman style packages, and was missing them.

From #423
>  Honestly. the apk packager is missing a lot of tests to achieve testing
> parity with deb and rpm.

I had used the apk implementation (and unit tests) as a reference, because the
package formats are similar.
I based those tests, and my view of what tests are missing on the apk packager.

> However, I'm not sure what the benefits of intentionally failing acceptance
> tests are. If broken packages can be produced we should already prevent
> that with unit tests. Acceptance tests should rather test if the features
> we support do actually work on a real system.

True, but personaly I think there is value in testing the error handeling.
But maybe the place for that would be `end to end tests`, rather than acceptance tests?

# Simple things:

> a empty folder: this one is already covered by the unit tests which check if
> the entry for a directory is created with the correct type mode and if it
> can be extracted again from the tar. Testing this again in an acceptance
> test would be more of a test of dpkg/apt than of nfpm.

okay, i don't see this tested in the apk packager.
For dpkg, one may add a `.placeholder` file to an empty directory.

This seems to be relevant while `apt purge`ing a package.

> Epoch is set: This mechanisms works identically to, for example, the version
> field which is already tested. There is some value to an acceptance test
> (checking if we spelled the field name correctly such that dpkg recognizes
> it) but acceptance tests are more expensive so I guess the unit tests are
> enough given that the mechanisms itself if tested in the acceptance tests
> for other fields.

For pacman, the epoch is included in the package file name, if it is set.
While building the logic for the filename, i was missing such a test case.

>a broken symlink: This is the responsibility of the user. Afaik nfpm allows
> for valid and invalid symlinks and we shouldn't change that because we don't
> know what the user wants to achieve with it. For example, an rpm could be
> built with a symlink to valid files in Fedora that just don't exist in
> CentOS.

Currently nfpm just adds the symlink, but at least for pacman packages this is
not allowed and results in an error.


> a path longer than 256 characters: If this violates cpio/tar spec nfpm
> should error out, then there is nothing to test. If some filesystems don't
> accept this, this is also in the responsibility of the user because users
> maybe build for distros with alternative file systems.

It does not vialate cpio/tar spec.
Most/all common unix filesystems allow paths up to `2^15` (32,767) characters long.


> A path with windows style seperators: This should also be validated in a
> unit test. I don't know what actually happens. If a file containing \ in
> the name is created we could warn the user but that also might be a false
> positive.

I will add an according unit test instead.


> a file with a leading space/A path with windows style seperators: Files are
> binary blobs and can contain anything. This would just test the Go standard
> library and dpkg/rpm.

Not sure this is clear: A file with a leading space in the filename, not the contents.  
ala `/usr/local/bin/ file`  
I added this specifically because i dont know how all the packagemanagers react.
I would argue that this should be tested, because its exaclty the annoying edgecase that always bites me

* If it works now, Great.
* If it doesnt work, we should document that.

> A path starting with a windows style C:/: We could handle that with an
> custom error but even then, this should be a unit test.

I will add a unit test.
For source paths, this may also just work<sup>tm<sup/>.



# File Validation tests

> a file that is included twice: Already tested in TestCollision in files_test.go.

Okay.

> a file with a name longer than 256 characters: See above

This is valid for as least the tar based packages, but the install may fail
because the underliying filesystem may or may not allow this.

> file with a space/leading space/newline/null: All of them are valid instead of
> the null byte. We should probably just strip null bytes in toNixPath in
> files.go and add a test in files_test.go.

Will add a unit test.

# Path validation tests

> A src path with: `..` Already tested in multiple tests.

Should nfpm allow going above the working directory?


> 5 different paths that are valid, but not sanitized: We could add some sanity
> checks in files.go and corresponding unit tests in files_test.go.

Will add unit tests.

> A dst path with ../.., thus escaping above /: There is no escaping the
> filesystem root afaik, but we already clean the destination anyway but we
> could still add a unit test in files_test.go.

This should either not be allowed, and produce an error,
Or should result in a path starting with `/`

> 5 different paths that are valid, but not sanitized: We could add some
> sanity checks in files.go and corresponding unit tests in files_test.go.

Okay.

> A dst path with ../.., thus escaping above /: There is no escaping the
> filesystem root afaik, but we already clean the destination anyway but we
> could still add a unit test in files_test.go.
 
Will add a unit test

----

Its now at unit tests for:

* Windows style seperators
* A windows style `C:\`
* paths with
  * newline
  * space
  * leading space
  * null character
  * A dst path with ../.., escaping above /
* Path sanitizaion tests

Acceptance tests for:

* Epoch set
* Broken symlink
* leading space.

All the common tests missing in the apk packager.

erikgeiser:
> True, but personaly I think there is value in testing the error handeling.
But maybe the place for that would be end to end tests, rather than acceptance tests?

Well the error handling happens during building and the build process is covered by the unit tests. If we want to make sure that we handle duplicate files gracefully, we add a unit test that checks the error (for this example, see `TestCollision` in `files_test.go`). The acceptance tests are to make sure that packages that are successfully built using `nfpm` do actually work as intended on different distros. An error case would mean that no package can be created and thus no acceptance test can run.

> okay, i don't see this tested in the apk packager.
For dpkg, one may add a .placeholder file to an empty directory.

That's because it is tested one level deeper in `files_test.go` which handles the content for all packagers.

> For pacman, the epoch is included in the package file name, if it is set.
While building the logic for the filename, i was missing such a test case.

We should add epoch tests for all packagers then, but in the unit tests.

> Currently nfpm just adds the symlink, but at least for pacman packages this is
not allowed and results in an error.

Is this the case for any broken symlink or for symlinks that don't point to a file in the same package? If it's the latter, `nfpm` can check for it in the packager and we can add a test that covers this check. If it's any broken symlink we can't do anything about because we don't know if the file it points to exists and the time of packaging.

>  It does not vialate cpio/tar spec.
Most/all common unix filesystems allow paths up to 2^15 (32,767) characters long.

Then I don't think this warrants a test.

> Not sure this is clear: A file with a leading space in the filename, not the contents.
ala /usr/local/bin/ file
I added this specifically because i dont know how all the packagemanagers react.
I would argue that this should be tested, because its exaclty the annoying edgecase that always bites me
>If it works now, Great.
 If it doesnt work, we should document that.

That's just a file name that happens to start with a space. Keep in mind, we don't have to test `rpm` and `apk` and so on. And we don't know that filesystem `rpm`, `dkpg`, ... may work on. A space in the file name is so common that I don't think it is something we have to test for.

> This is valid for as least the tar based packages, but the install may fail
because the underliying filesystem may or may not allow this.

Yeah but we can't do anything about that. Only the users know which distros and which filesystem they want to target.

> This should either not be allowed, and produce an error,
Or should result in a path starting with /

It results in a path starting with `/` because in `files.go` we clean the path and convert the separators to `/`. Cleaning `/etc/../../../../foo` results in `/foo`, just like when you do it manually using `cd` for example.

4censord:
Thank you, you helped me quite a bit.

I will be back with more unit tests

erikgeiser:
Glad have helped. I'm sorry that some of my suggestions seem a bit nit-picky but we do definitely appreciate any contributions. Even just a discussion on how we conduct tests (or how we can conduct them in the future) is always appreciated.

4censord:
I just found 2 more:

* is `config|noreplace` honored
* are files marked for a different packager actually ignored?

PR #431, opened 2021-12-28
fix: create tree of dir content types

caarlos0:
refs https://github.com/goreleaser/goreleaser/issues/2792

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -414,6 +414,10 @@ func createFilesInsideTarGz(info *nfpm.Info, tw *tar.Writer, created map[string]
 			continue
 		}
 
+		if err := createTree(tw, file.Destination, created); err != nil {
+			return err
+		}
+
 		normalizedName := normalizePath(strings.Trim(file.Destination, "/")) + "/"
 
 		if created[normalizedName] {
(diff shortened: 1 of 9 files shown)


erikgeiser:
This solution assumes that the contents are sorted because by creating the parent tree of `dir` you rely on the fact that none of the following `dir` entries (that possibly have different permissions, etc.) are already including in the parent tree of the earlier entry.

Now we do sort the contents, but in a completely different package. Keep in mind that just assuming order in the packagers might be a huge foot gun for those that use `nfpm` as a library.

caarlos0:
indeed, should we maybe sort the files in a different place then (in each impl)?

or add a godoc saying "if you use this directly you'll need to sort the files first"?

thoughts?

erikgeiser:
Well we already create the tree of the `dir` entries already, just in the next loop.

I think the best solution would require some refactoring: We should build the actual tar contents first without writing them. Then each content entry can decide if it should overwrite existing entries or not. Like `createTree` would do nothing if an directory already exists and a `dir` entry would override an existing dir that may have been created by `createTree` before. Then we write it all afterwards. In this case order would be just cosmetics.

caarlos0:
that seems like a rather big refactor... I think I'll release this as-is to fix the bug, and create an issue so we can investigate it in the future

Issue #439, opened 2022-01-31
check debian packages with lintian


caarlos0:
we should probable check deb packages with [lintian](https://lintian.debian.org) in the unit/acceptance tests to make sure they are ok.

refs https://github.com/goreleaser/goreleaser/issues/2861

caarlos0:
not really possibly as it always complains that we use static binaries... 

some of the issues are fixable, though, will probably add a note to the docs only.

PR #445, opened 2022-02-03
fix: lintian issues

caarlos0:
refs #439

this fixes some of lintian's issues with nfpm package itself.


CryptoGodfatherVA2 (commented):
Var5

PR #454, opened 2022-02-09
Expand dependencies' version

Felixoid:
Now the version is expanded in version, but in case, when many packages are built and depend on near one with the same version it doesn't work.


Felixoid:
Here's an example of a problem https://github.com/ClickHouse/ClickHouse/blob/8d475e8c135c311b5a81e059b680eeee17b3bb5c/packages/clickhouse-client.yaml#L29

The packages are built together, but VERSION is expanded only partially.

djgilcrease (approved):
LGTM

Felixoid:
is `docs / htmltest` test something that I should care about?

caarlos0:
> is `docs / htmltest` test something that I should care about?

no idea why it started to fail recently, will investigate eventually

caarlos0 (approved):
looks good, thanks!

Felixoid:
Thanks!

PR #456, opened 2022-02-12
feat: sign our packages

caarlos0:
this should sign our packages.

apk signing is kind of useless, as it will still complain that the package signature is untrusted, even if you put the public key in `/etc/apk/keys`. 

I think it expects you to create your own apk repository (`APKINDEX.tgz`) and sign these as well. Maybe I'm wrong though, happy to hear your opinions.

If this is acceptable, I'll push the public keys to our `.github` repository, and update the docs. Will use the same keys for goreleaser itself.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,11 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - if: success()
+        name: create-keys
+        run: |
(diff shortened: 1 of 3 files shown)


kaniini:
> I think it expects you to create your own apk repository (APKINDEX.tgz) and sign these as well. Maybe I'm wrong though, happy to hear your opinions.

No, if the key is in /etc/apk/keys and named `foo.rsa.pub` where the signing key is named `foo.rsa`, it should work.

djgilcrease (approved):
LGTM

PR #458, opened 2022-02-12
fix: update rpmpack and change default compressor to gzip:-1 (a default level) instead of gzip:9

Felixoid:
The default compression level for rpm is 9. https://github.com/google/rpmpack/blob/master/rpm.go#L165

Because of it, rpm packaging is significantly slower than deb, almost 5 times worse.

Here I bump the dependencies, add tests and change default level to -1


caarlos0:
I think we can leave the default as `9`,  and users who feel it is too slow can change it. WDYT?

Felixoid:
The difference between 5 and 9 levels in speed is huge, 1.5 vs 7.5 minutes, and only 5% in space. I'd say, 9 is rather worthless.

caarlos0:
thanks!

Felixoid:
Thank you for reviewing!

BTW, the bot just added 2.12.0 milestone. Looks like something wrong since 2.13 is released.

PR #457, opened 2022-02-12
feat: replace compress/gzip by klauspost/pgzip

Felixoid:
Here's the difference between compress/gzip and https://github.com/klauspost/pgzip binary on 16 cores laptop

```
Feb 12 19:15:42 + nfpm package --target /output --config clickhouse-common-static.yaml --packager deb
Feb 12 19:15:42 using deb packager...
Feb 12 19:17:15 created package: /output/clickhouse-common-static_22.2.1.1638_amd64.deb
93 seconds

Feb 12 19:34:00 + ./nfpm package --target /output --config clickhouse-common-static.yaml --packager deb
Feb 12 19:34:00 using deb packager...
Feb 12 19:34:13 created package: /output/clickhouse-common-static_22.2.1.1638_amd64.deb
13 seconds
```

The resulting deb package is 1015M.

I am leaving a standard `compress/gzip` in tests to have a reference.

PR #475, opened 2022-03-14
Update rpmpack to version with pgzip

Felixoid:
As the following up for https://github.com/google/rpmpack/pull/72 and #457 , here's an update for rpm pgzip


djgilcrease (approved):
LGTM

caarlos0:
likely over the weekend

PR #491, opened 2022-04-06
feat: add support for additional fields in Debian control files

### LINKED ISSUE #490: Add support for arbitrary .deb control file fields

Debian control files [have a number of optional fields](https://man7.org/linux/man-pages/man5/deb-control.5.html) -- including "Bugs" and "Tags" -- that are used by many existing packages. The original `fpm` [supports adding arbitrary fields to Debian control files with the --deb-field flag](https://fpm.readthedocs.io/en/latest/packages/deb.html?highlight=changelog#deb-specific-command-line-flags), but in my understanding it's not currently possible to do so with `nfpm`.

I'm currently trying to port some packaging code from `fpm` and `nfpm` and it seems unfortunate to give up these fields to do so.

I'd like to see support for custom Debian control fields added to `nfpm`. My imagined configuration would add a new setting under the existing `deb` section like this:

```yaml
deb:
  fields:
    - name: Bugs
      value: https://github.com/goreleaser/nfpm/issues
```

Since these fields are optional, `nfpm` may chose not to support them. As I mentioned, I think this would be unfortunate and keep folks on `fpm` in some cases.

### Search

### Code of Conduct

### Additional context

I'd be willing to implement this if there is interest.

atheriel:
Debian control files (DCF) have [a number of optional fields](https://man7.org/linux/man-pages/man5/deb-control.5.html) -- including "Bugs" and "Tags" -- that are used by many existing packages. The original `fpm` supports adding arbitrary fields to DCFs [with the `--deb-field` flag](https://fpm.readthedocs.io/en/latest/packages/deb.html?highlight=changelog#deb-specific-command-line-flags), but until now it was not possible to do so with `nfpm`.

This PR introduces a new deb-specific override, `fields`, that simply translates key-value pairs from a map directly into the DCF.
Given the following `nfpm` config file entry:

    deb:
      fields:
        Bugs: https://github.com/goreleaser/nfpm/issues

The resulting DCF would contain the entry

    Bugs: https://github.com/goreleaser/nfpm/issues

To avoid generating malformed DCFs, we simply ignore empty entries.

This PR includes unit tests and updates to the website documentation.

Closes #490.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -718,6 +718,11 @@ Homepage: {{.Info.Homepage}}
 {{- end }}
 {{- /* Mandatory fields */}}
 Description: {{multiline .Info.Description}}
+{{- range $key, $value := .Info.Deb.Fields }}
+{{- if $value }}
+{{$key}}: {{$value}}
+{{- end }}
+{{- end }}
 `
 
 type controlData struct {
--- a/deb/deb_test.go
+++ b/deb/deb_test.go
@@ -1267,3 +1267,34 @@ func TestArches(t *testing.T) {
 		require.Equal(t, "foo64", info.Arch)
 	})
 }
+
+func TestFields(t *testing.T) {
+	var w bytes.Buffer
+	require.NoError(t, writeControl(&w, controlData{
+		Info: nfpm.WithDefaults(&nfpm.Info{
+			Name:        "foo",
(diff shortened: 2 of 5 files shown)


djgilcrease (approved):
LGTM

PR #375, opened 2021-10-09
fix: deb release at end of version string

kpenfound:
Fix for https://github.com/goreleaser/nfpm/issues/369

This moves `info.Release` to the end of the version string, effectively applying `info.Prerelease` and `info.Metadata` to the upstream version rather than the debian-revision

Effective change to the deb filename is:
Before:
- Version: `1.2.3~beta+abc`
- Release: `2`
- Output filename: `foo_1.2.3-2~beta+abc_amd64.deb`

After
- Version: `1.2.3~beta+abc`
- Release: `2`
- Output filename: `foo_1.2.3~beta+abc-2_amd64.deb`

PR #498, opened 2022-04-22
fix: deb version order in control

kpenfound:
This should have been caught in https://github.com/goreleaser/nfpm/pull/375 but I missed it.

The version string when using `Release` is correct in the filename, but incorrect in the control file.

Here's the [debian docs on the control file's version](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version)

Before:
Filename: `foo_1.2.3~pre+meta-2_amd64.deb`
control version: `1.2.3-2~pre+meta`

After:
Filename: `foo_1.2.3~pre+meta-2_amd64.deb`
control version: `1.2.3~pre+meta-2`

✅ `task ci` 🚀


caarlos0:
thanks!

PR #508, opened 2022-05-16
feat: implement dpkg-sig Package signing

SasSwart:
As mentioned [here](https://github.com/goreleaser/nfpm/issues/212) Debsign package verification is a PITA.

Herewith my proposal to implement dpkg-sig style package verification.
As mentioned in the original thread, debsign and dpkg-sig are mutually exclusive, but it would be nice to be able to choose between them.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -219,7 +219,7 @@ func newDpkgSigFileLine(name string, fileContent []byte) dpkgSigFileLine {
 	}
 }
 
-func readDpkgSigData(info *nfpm.Info, debianBinary, controlTarGz, dataTarball []byte) io.Reader {
+func readDpkgSigData(info *nfpm.Info, debianBinary, controlTarGz, dataTarball []byte) (io.Reader, error) {
 	data := dpkgSigData{
 		Files: []dpkgSigFileLine{
 			newDpkgSigFileLine("debian-binary", debianBinary),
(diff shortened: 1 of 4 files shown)


SasSwart:
I noticed an issue with how I'm populating the template.
I'm converting the PR to draft for the moment so I can fix that.

SasSwart:
Template issue fixed.
Ready for review

caarlos0:
Hey! Thanks for the pr! I wonder if we can leave only the new method and remove the deprecated one? Does latest debian supports the new one already? Any major distribution that doesn't?

SasSwart:
I have fixed the linting issue and added some more configs. All tests, linting etc. should now pass.

I've also done some more reading and I am convinced that the other signing method will have to be kept.
It still is the default method supported by dpkg.
That being said, I'm pretty sure neither are used much.

Debian/Apt seem to depend far more on signing the Release files in their repositories than they depend on signing packages themselves. Packages are usually only signed when distributed without a repository. Even in those cases, the big boys like Microsoft seem not to bother with signing the package that is available for download. See https://code.visualstudio.com/ for an example.

Debian's package building depends pretty heavily on Source packages, which have their own detached control file (called a .dsc file) which is signed. None of this is as yet supported by nfpm.

I have seen another issue in this repository about supporting Debian source packages. The response seems to have been that you do not think that belongs within the purview of nfpm.

I disagree with this. I think there is a way that we can reasonably support building source packages. I'll attempt a PR and submit one if its possible.

My source for most of the above:
https://debian-handbook.info/browse/stable/sect.source-package-structure.html

caarlos0:
> I have seen another issue in this repository about supporting Debian source packages. The response seems to have been that you do not think that belongs within the purview of nfpm.

I think it was rpm source packages... tbh I'm not sure where it belongs, or if we should implement it...

caarlos0:
Thanks, adding some stuff on #515 if you wanna check it out

Issue #474, opened 2022-03-09
apk version strings are incorrect


kaniini:
### What happened?

In https://github.com/goreleaser/nfpm/blob/main/apk/apk.go#L562-L565, the version format is Debian-like.  APK uses [Gentoo-style versioning](https://wiki.gentoo.org/wiki/Version_specifier) and will not understand these versions correctly.

### How can we reproduce this?

Generate any APK with prerelease or epoch set.  It will generate a bogus version string.

### nfpm version

```bash
whatever git master is
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

It would also be nice to refactor the apk writing part into its own module.  At present, I cannot use nfpm's apk module for my own purposes :(

caarlos0:
hey @kaniini , thanks for bringing this up.

if you want to work on a PR for this it would be amazing... I can help through here/slack/discord if you need :)

PR #514, opened 2022-05-26
Fix apk versioning

kpenfound:
Fixes https://github.com/goreleaser/nfpm/issues/474

Following examples from https://wiki.gentoo.org/wiki/Version_specifier as well as https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgver

I used `Info.Prerelease` as the `pkgver.suf` as described above, since that maps most similarly to the semver prerelease field. It potentially has some overlap with `Info.VersionMetadata` though. For parity with these fields for other package types in nfpm, it would maybe make sense to append both Prerelease and VersionMetadata to the version, however we would have to ensure that only one of the two fields is used.

Maybe it's better to expect the user to put  `{letter}` and `{suf{#}}` in `Info.Version` instead.

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -90,16 +90,13 @@ type Apk struct{}
 func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
 	info = ensureValidArch(info)
 	version := info.Version
-	if info.Release != "" {
-		version += "-" + info.Release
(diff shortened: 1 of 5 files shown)


caarlos0:
LGTM, thanks!

PR #515, opened 2022-05-30
feat: implement dpkg-sig Package signing

caarlos0:
continuing #508


erikgeiser (approved):
LGTM. This PR doesn't contain unit tests for the new signing method though. I don't think that's a problem because the PR does add acceptance tests but I like the unit tests because they are way faster and more convenient. But overall it's not a big deal.

caarlos0:
@erikgeiser the unit tests were added in #508 

main issue in this pr imo is that it does not actually verify the signing, as I couldn't find any way of doing it with a custom signature...

maybe @SasSwart knows a way?

PR #533, opened 2022-07-27
fix: json schema

caarlos0:
the jsonschema lib was updated, and now only supports json tags...

this duplicates the json and yaml tags, which is horrible and I'm happy to hear if there's a better way...

https://github.com/invopop/jsonschema/pull/23


erikgeiser (approved):
I don't think there is a better way. I'd do it the same way.

djgilcrease (approved):
LGTM

PR #537, opened 2022-08-10
feat: removing changelog generation

caarlos0:
This reverts commit 47d15255b00e959c6626f81b33307adc0c7de547.  

refs https://github.com/goreleaser/goreleaser/issues/3259


caarlos0:
this looks horrible tbh, need to find a middle ground between the "magic" from before and something users can easily provide... no idea what that should look like yet, maybe the formatting stuff from chglog, without actually generating the changelog ... and maybe something simpler too... will sleep and think about it in the next days

PR #543, opened 2022-09-11
feat: add support for Arch Linux packages

### LINKED ISSUE #133: Support for Arch Linux packages?

### LINKED ISSUE #546: Handle packaging formats with non-distinct file extensions

When implementing #543, I imported it into a local clone of goreleaser in order to test it. I noticed that it created a file with the `.archlinux` extension. The issue is that the packager was registered as "archlinux", which I did because Arch Linux doesn't have a distinct package extension, and just uses `.tar.zst`, which is the extension for any TAR file compressed with Zstandard.

I think the best solution would be to add a new interface that looks something like this:

```go
type PackagerWithExtension interface {
	Packager
	ConventionalExtension() string // Returns an extension like .rpm or .pkg.tar.zst
}
```

This way, a breaking change is avoided, while providing a type-safe way to check the extension. To check the extension, you'd do something like

```go
if packager, ok := packager.(nfpm.PackagerWithExtension); ok {
	ext := packager.ConventionalExtension()
	// do something with ext
}
```

The `ConventionalExtension()` method could be added directly to the `Packager` interface, but that would be a breaking change. Also, an extension argument could be added to `RegisterPackager()`, but that would be a breaking change as well.

### Search

### Code of Conduct

### Additional context

_No response_

Arsen6331:
This PR adds support for Arch Linux packages.

Like the previous PR (#141), this one generates `.PKGINFO`, `.INSTALL`, and `.MTREE` files.

Unlike that PR, this one compresses the package with a pure-Go Zstandard implementation, which conforms with the new standard for Arch packages.

I have successfully been able to install packages generated by this code, so it works, but there are no tests, so I am leaving it as a draft until I add tests and more comments to it.

Closes #133 
Closes #546


caarlos0:
amazing @Arsen6331! Let me know if you need any help!

Arsen6331:
I've added the tests and comments, and fixed what the linter suggested.

caarlos0:
Thanks, @Arsen6331 🚀 

could you make the changes to the docs as well? they live in the `www` folder.

Arsen6331:
> could you make the changes to the docs as well? they live in the www folder.

Sure

Arsen6331:
I've changed the docs

arch/arch.go
```diff
@@ -0,0 +1,658 @@
...
+		info.Name,
+		info.Version,
+		pkgrel,
+		info.Arch,
+	)
+
+	return validPkgName(name)
+}
+
+// validPkgName remoces any invalid characters from a string
```
djgilcrease, line 80:
```suggestion
// validPkgName removes any invalid characters from a string
```

djgilcrease (changes requested):

djgilcrease (changes requested):
- [x] Implement signed packages https://wiki.archlinux.org/title/DeveloperWiki:Package_signing Appears to create a `.sig` file to live outside the package?
- [ ] Implement `acceptance` tests

Arsen6331:
> Implement signed packages [wiki.archlinux.org/title/DeveloperWiki:Package_signing](https://wiki.archlinux.org/title/DeveloperWiki:Package_signing) Appears to create a .sig file to live outside the package?

Yeah, it requires a separate file and I don't know how to implement that with the current API.

djgilcrease:
> 
> Yeah, it requires a separate file and I don't know how to implement that with the current API.

Just make a `doSign` method that writes the file alongside the package if the `info.ArchLinux.Signature.KeyFile` is not empty.

Arsen6331:
> Just make a `doSign` method that writes the file alongside the package if the `info.ArchLinux.Signature.KeyFile` is not empty.

The issue with that is that the `Package()` method of the `Packager` interface I've implemented takes an `io.Writer` and writes to it. The packager doesn't actually control file creation.

caarlos0:
> > Just make a `doSign` method that writes the file alongside the package if the `info.ArchLinux.Signature.KeyFile` is not empty.
> 
> The issue with that is that the `Package()` method of the `Packager` interface takes an `io.Writer` and writes to it. The packager doesn't actually control file creation.

ah yes, seems like one more thing we'll have to change... 

that said, maybe its for the best to actually break API compat for both this and #546?

caarlos0:
> I am ok skipping signing for now so we can take a little time and design the build/sign API.

yeah, sounds good

Arsen6331:
I've looked at the existing acceptance tests and I understand what they do, but I have no idea where to start in terms of implementing them for Arch. Can someone point me in the right direction?

caarlos0:
@Arsen6331 If I'm not missing anything:

- add arch here: https://github.com/goreleaser/nfpm/blob/e6f966fa6f3d94c5fddb4decc14aa9294a4f6d8a/acceptance_test.go#L22
- create the dockerfiles and edit the goreleaser config files here: https://github.com/goreleaser/nfpm/tree/main/testdata/acceptance

you can start with the `min` and `simple` to get a handle on how it works... it is a bit complex - mainly because properly testing all this is complex, but once you get a hand of it it will hopefully make more sense...

Arsen6331:
On Arch, packages cannot write to `/usr/sbin` as it is a symlink to `/usr/bin`, and pacman does not follow the symlink and therefore refuses to write to it. This is intentional behavior and not a bug in pacman. The `complex` acceptance test tries to add a file to `/usr/sbin`. How should I handle this? I could add an if statement in `TestCore()` that uses an Arch-specific file for the complex test. Would that be an acceptable solution?

Arsen6331:
Oh, I just realized I could also do this if it's better:

```yaml
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: deb
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: rpm
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: apk
- src: ./testdata/fake
  dst: /usr/bin/fake
  file_info:
    mode: 04755
  packager: archlinux
```

I think both solutions are a bit ugly, but I don't see any other solution.

caarlos0:
I don't remember why we write to sbin to be honest, maybe someone else does? cc\ @goreleaser/nfpm 

if not, imho, we can change it to `/usr/bin`. 

if there's a reason, I would probably create a `complex` test config just for arch... probably...

djgilcrease:
I kind of like the complex config for the complex test as it shows what you may need to do

```yaml
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: deb
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: rpm
- src: ./testdata/fake
  dst: /usr/sbin/fake
  file_info:
    mode: 04755
  packager: apk
- src: ./testdata/fake
  dst: /usr/bin/fake
  file_info:
    mode: 04755
  packager: archlinux
```

Arsen6331:
There is this multi-architecture image: [lopsided/archlinux](https://hub.docker.com/r/lopsided/archlinux/), which supports `amd64` and all the ARM variants, but it doesn't contain things like `grep`, which are used by the acceptance tests.

Arsen6331:
There also seems to be a `devel` tag on that image, which has the `base-devel` group installed, so it has all the tools I need. That one should work, but then there's still no way to test `386`.

caarlos0:
> I've implemented the acceptance tests. I added only `amd64` to `acceptance_test.go` for Arch because Arch only officially supports `x86_64`. This means that the official docker image only works on `amd64`. However, we should still support the unofficial architectures because they're used very often, especially the armv7 and arm64 ones. I don't know how to proceed with that.

I would say that, at least for now, we can go with the official support only...

Arsen6331:
Ok, in that case, the acceptance tests are finished and pass. I did test them locally for both amd64 and the ARM variants, and they all pass there too.

Arsen6331:
Should I add `archlinux` to the `pkgFormat` field in `.github/workflows/build.yml`?

caarlos0:
> Should I add `archlinux` to the `pkgFormat` field in `.github/workflows/build.yml`?

yes, please!



> Ok, in that case, the acceptance tests are finished and pass. I did test them locally for both amd64 and the ARM variants, and they all pass there too.

awesome!

Arsen6331:
Done

djgilcrease (approved):
LGTM

Arsen6331:
I could also try to make a PR against https://github.com/goreleaser/goreleaser that integrates these changes and can be merged once they make it into a release.

apk/apk.go
```diff
@@ -101,6 +101,11 @@ func (a *Apk) ConventionalFileName(info *nfpm.Info) string {
 	return fmt.Sprintf("%s_%s_%s.apk", info.Name, version, info.Arch)
 }
 
+// ConventionalExtension returns the file name conventionally used for Apk packages
+func (*Apk) ConventionalExtension() string {
+	return ".deb"
```
caarlos0, line 106: should be apk here

caarlos0 (commented):
on a first look it looks good to me! will see if I find some arch contributor to help review it as well! 

thanks again for the work on this! ❤️

caarlos0:
> I could also try to make a PR against https://github.com/goreleaser/goreleaser that integrates these changes and can be merged once they make it into a release.

if you feel like it, go for it! If not, I can do it as well, no issues :)

apk/apk.go
Arsen6331: Oh, you're right, one second

Arsen6331:
I went through a bunch of PKGBUILDs on the AUR to see if I missed anything and implemented all of it.

caarlos0:
I'll review this again today and see that we have it merge soon! 🙏 Sorry for the delay, had a crazy couple of weeks :P

Arsen6331:
> I'll review this again today and see that we have it merge soon! pray Sorry for the delay, had a crazy couple of weeks :P

No problem. Merge it whenever it's convenient for you, there's no rush.

azat:
Apparently right now archlinux packager does not support `type=config` for directories:

```
nfpm package --target .. --config clickhouse-server.yaml --packager archlinux
using archlinux packager...
read root/etc/clickhouse-server: is a directory
```

caarlos0:
> Apparently right now archlinux packager does not support `type=config` for directories:
> 
> ```
> nfpm package --target .. --config clickhouse-server.yaml --packager archlinux
> using archlinux packager...
> read root/etc/clickhouse-server: is a directory
> ```

can you open an issue/PR?

azat:
>can you open an issue/PR?

#564

Issue #501, opened 2022-05-04
Post-remove script example seems incorrect


gregcotten:
### What happened?

This [example in the docs](https://github.com/goreleaser/nfpm/blob/1f50ccc8302150320552d7fe6dd9e2c15d604a1e/www/docs/tips.md#example-multi-platform-rpm--deb-post-remove-script) seems to not actually be an example of post-remove, as it references install + upgrade.

### How can we reproduce this?

View the URL

### nfpm version

```bash
n/a
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Dean-Coakley:
@caarlos0 I was considering proposing a fix of just replacing the example with the contents of https://github.com/goreleaser/nfpm/blob/main/testdata/acceptance/scripts/postremove.sh

Would that be sufficient?

djgilcrease:
yup that would work

PR #549, opened 2022-09-23
Fix post-remove script example

Dean-Coakley:
Fixes: https://github.com/goreleaser/nfpm/issues/501

Fix post-remove example in docs. Based on https://github.com/goreleaser/nfpm/blob/main/testdata/acceptance/scripts/postremove.sh


djgilcrease (approved):
LGTM

PR #567, opened 2022-10-26
Add Prefix option to create relocatable packages.

scr-oath:
In order to create "relocatable" packages, one may provide the Prefix tag, which marks the part to be stripped off when relocating.

- See https://ftp.osuosl.org/pub/rpm/max-rpm/s1-rpm-inside-tags.html#S3-RPM-INSIDE-PREFIX-TAG


scr-oath:
Hmm… http://ftp.rpm.org/api/4.4.2.2/relocatable.html shows the need for multiple prefixes - not ONLY one.  I'll have to rethink this… marking draft for now

PR #568, opened 2022-10-27
Allow using directories as contents in archlinux packager

### LINKED ISSUE #564: archlinux packager does not support folders for configs
### What happened?

Tried to create archpackage with the following configuration and failed:

```yaml
$ cat foo.yaml 
name: test

contents:
- src: etc/foo
  dst: /etc/foo
  type: config
```

```sh
mkdir -p etc/foo
touch etc/foo/foo.yml
```

```
$ nfpm package --target . --config foo.yaml --packager archlinux
using archlinux packager...
read etc/foo: is a directory
```

Though debian is OK:

```
$ nfpm package --target . --config foo.yaml --packager deb
using deb packager...
DEPRECATION WARNING: Leaving the 'maintainer' field unset will not be allowed in a future version
created package: test_0.0.0~rc0_amd64.deb
```

### nfpm version

```bash
nfpm version nfpm version v2.20.0
```

### Search

### Code of Conduct

### Additional context

_No response_

Arsen6331:
This PR allows using directories in the contents when using the archlinux packager. It also adds acceptance tests for all the packagers so that no one misses this use case in the future.

Fixes #564


caarlos0:
noice, thanks @Arsen6331 !

PR #576, opened 2022-11-13
fix: relative paths to dirs

### LINKED ISSUE #575: Confusing behaviour with trailing `/` in destination
### What happened?

I have the following nfpm.yaml:

```yml
name: foo
version: 0.0.0
maintainer: asdf
arch: amd64
contents:
- dst: /usr/local/share/file/
  src: file.txt
```

Note the `/` at the end of the destination path. Because of that, I expect nfpm to install the file as `/usr/local/share/file/file.txt`. However, it installs the file as `/usr/local/share/file`. This is confusing because the trailing `/` indicates that the path is a directory, so the logical course of action would be to put the file into the directory.

Copy the above nfpm.yaml, and create a file `file.txt`, and run `nfpm package -p deb`

### nfpm version

```bash
nfpm version nfpm version 2.22.0
commit: 5c3cc38943f1743bcaeb422c2dc65555fc9b12a1
built at: 2022-11-12T03:50:16Z
built by: goreleaser
module version: v2.22.0, checksum: h1:TC6tuxxFq0hL0WapI63+CkLMbcsYRbR7kAnLkRYAEbo=
```

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
closes #575

### DIFF
--- a/files/files_test.go
+++ b/files/files_test.go
@@ -370,3 +370,15 @@ func TestGlobbingFilesWithDifferentSizesWithFileInfo(t *testing.T) {
 		t.Fatal("test FileInfos have the same size, expected different")
 	}
 }
+
+func TestDestEndsWithSlash(t *testing.T) {
+	result, err := files.ExpandContentGlobs(files.Contents{
+		{
+			Source:      "./testdata/globtest/a.txt",
+			Destination: "./foo/",
+		},
+	}, false)
+	require.NoError(t, err)
+	require.Len(t, result, 1)
+	require.Equal(t, "foo/a.txt", result[0].Destination)
+}
--- a/internal/glob/glob.go
+++ b/internal/glob/glob.go
@@ -92,6 +92,11 @@ func Glob(pattern, dst string, ignoreMatchers bool) (map[string]string, error) {
 			continue
 		}
 
+		if strings.HasSuffix(dst, "/") {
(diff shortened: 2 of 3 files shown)


djgilcrease (approved):
LGTM

drakkan:
Hi,

this change broke my use case.

Please take a look [here](https://github.com/drakkan/sftpgo/blob/main/pkgs/build.sh#L77). 
I now have this error:

```
cannot add "/home/nicola/goprojects/sftpgo/templates/webclient/twofactor.html" because "/home/nicola/goprojects/sftpgo/templates/webadmin/twofactor.html" is already present at the same destination (/usr/share/sftpgo/templates/twofactor.html): content collision
```

I have some files with the same name in `templates/webclient` and `templates/webadmin`. Can I fix this by changing my glob syntax? (I've tried some alternate globs with no luck). Thanks

caarlos0:
@drakkan yes, unfortunately this might be a breaking change for some... but I still think it was a bug before... I'll add a note about it in the readme though.

caarlos0:
https://github.com/goreleaser/nfpm/releases/tag/v2.22.1

drakkan:
> @drakkan yes, unfortunately this might be a breaking change for some... but I still think it was a bug before... I'll add a note about it in the release notes though.

no problem, thanks

PR #579, opened 2022-11-18
docs: Add more detail to configuration docs

j-tai:
This adds more detail to the configuration reference in the docs. The primary goal is to help users understand what each field does even if they have not created packages before.


caarlos0:
thank you!

PR #581, opened 2022-11-21
fix: improve contents.type

caarlos0:
improved jsonschema and added validations

### DIFF
--- a/arch/arch.go
+++ b/arch/arch.go
@@ -477,7 +477,7 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 	}
 
 	for _, content := range info.Contents {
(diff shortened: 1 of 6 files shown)


arch/arch.go
```diff
@@ -477,7 +477,7 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 	}
 
 	for _, content := range info.Contents {
-		if content.Type == "config" || content.Type == "config|noreplace" {
```
djgilcrease, line 480: Are you sure you want to remove the `config|noreplace` here? It's is documented to act just like config except on rpm which does something special with it

caarlos0: hmmm I think deb also do not have it... I'll add it back, add tests, and add it to the other formats too

PR #584, opened 2022-12-01
docs: Discourage "v" prefix for version numbers

j-tai:
Using a `v` prefix for the `version` field happens to work if the version is semver-compatible, since the semver parser will strip the `v`. If it is not semver-compatible, then a `v` prefix should not be used, since some package managers (like deb) won't accept it.


caarlos0:
thanks!

Issue #589, opened 2022-12-13
Env expansion not working for global replaces, conflicts


mayocream:
### What happened?

Env expansion introduced in #548 only works for deb.Overrides, but not working for global replaces, conflicts.

### How can we reproduce this?

Example config file:
```yaml
name: "mayo"
arch: ${ARCH}
platform: "linux"
version: 1.0.0
section: "default"
priority: "extra"
maintainer: "Mayo"
vendor: "Mayo Inc."
license: "ASL 2.0"
replaces:
  - foobar
  - ${REPLACE_BLA}
conflicts:
  - mercurial
  - ${CONFLICTS_BLA}
contents:
- src: /tmp/mayo
  dst: /usr/local/bin
```

Result:
```bash
$ REPLACE_BLA=mayo nfpm pkg -f tests/nfpm.yaml -p deb
new Debian package, version 2.0.
size 740 bytes: control archive=386 bytes.
     1 bytes,     0 lines      conffiles            
   223 bytes,    10 lines      control              
    50 bytes,     1 lines      md5sums              
Package: mayo
Version: 1.0.0
Section: default
Priority: extra
Architecture: amd64
Maintainer: Mayo
Installed-Size: 0
Replaces: foobar, ${REPLACE_BLA}
Conflicts: mercurial, ${CONFLICTS_BLA}
Description: no description given
```

### nfpm version

```bash
nfpm version nfpm version 2.22.2
commit: 66760061ee255a163cbc77c7193826c15837fc0a
built at: 2022-11-30T12:46:25Z
built by: goreleaser
module version: v2.22.2, checksum: h1:K/how1xpNJC4xHpgMKQwE46iJp8+h0iTxWhhMYizhAA=
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

https://github.com/goreleaser/nfpm/blob/c80d81b47bca5161a86157fe88a154ab4aa0d2a8/www/docs/configuration.md?plain=1#L107-L112

mayocream:
> Will get to this tomorrow

Just wonder is it feasible to make env expansion also work on `name`? Thanks!

PR #593, opened 2022-12-14
Fix expansion of env vars in non-overriden package relation fields

djgilcrease:
closes: https://github.com/goreleaser/nfpm/issues/589


caarlos0 (commented):
looks good, should we add tests for this?

PR #596, opened 2022-12-30
fix: do not encourage using /usr/local/bin

caarlos0:
Examples, tests and docs often refer to /usr/local/bin, which should only be used for things compiled locally, which is not the case of installed packages.

Changed that to /usr/bin, and also some usages of things like /usr/something to /usr/share/something.

### DIFF
--- a/arch/arch_test.go
+++ b/arch/arch_test.go
@@ -42,7 +42,7 @@ func exampleInfo() *nfpm.Info {
 			Contents: []*files.Content{
 				{
 					Source:      "../testdata/fake",
-					Destination: "/usr/local/bin/fake",
+					Destination: "/usr/share/bin/fake",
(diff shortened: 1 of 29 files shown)


arch/arch_test.go
```diff
@@ -42,7 +42,7 @@ func exampleInfo() *nfpm.Info {
 			Contents: []*files.Content{
 				{
 					Source:      "../testdata/fake",
-					Destination: "/usr/local/bin/fake",
+					Destination: "/usr/share/bin/fake",
```
atc0005, line 45: @caarlos0 Was this meant to be `/usr/bin/fake` ?

caarlos0: yes, good catch, fixed in 475f856

PR #601, opened 2023-01-06
feat: expand variables in platform field

### LINKED ISSUE #600: Platform field variable expansion not happening
### What happened?

I was creating packages, using a template with some fields:

```
name: "nerdctld"
arch: "${GOARCH}"
platform: "${GOOS}"
version: "${VERSION}"
...
```

Everything looked fine, but it turns out the rpm was invalid:

```
package nerdctld-0:0.2.1-1.x86_64 is intended for a different operating system
```

rpm --xml shows why:
```xml
  <rpmTag name="Os">
	<string>${GOOS}</string>
  </rpmTag>
  <rpmTag name="Arch">
	<string>x86_64</string>
  </rpmTag>
```

The deb package is fine, seems like it doesn't even use platform ?

(it is supposed to go in the `Architecture` field for deb, like darwin-amd64, default is linux)

Try to use environment variables for platform and arch...
See that they are only expanded for arch, not for platform.

### nfpm version

```bash
nfpm version nfpm version 2.23.0
```

### Search

### Code of Conduct

### Additional context

My workaround is to comment out the "platform", for now.

```
arch: "${GOARCH}"
#platform: "${GOOS}"
version: "${VERSION}"
```

Since the software I am packaging is linux-only, anyway...

But I can submit a test case and a patch, simple one-liner.

afbjorklund:
Surprising that GOARCH was expanded, but GOOS was not.
More consistent to expand both of them the same way ?

Implement platform for deb packager, goes into Architecture.    
Validate that the platform is "linux", for Alpine and Arch.

Closes #600

### DIFF
--- a/nfpm_test.go
+++ b/nfpm_test.go
@@ -244,12 +244,28 @@ func TestOptionsFromEnvironment(t *testing.T) {
 		debPass         = "password123"
 		rpmPass         = "secret"
 		apkPass         = "foobar"
+		platform        = "linux"
+		arch            = "amd64"
 		release         = "3"
 		version         = "1.0.0"
 		vendor          = "GoReleaser"
 		packager        = "nope"
 		maintainerEmail = "nope@example.com"
 	)
+
+	t.Run("platform", func(t *testing.T) {
+		t.Setenv("OS", platform)
+		info, err := nfpm.Parse(strings.NewReader("name: foo\nplatform: $OS"))
+		require.NoError(t, err)
+		require.Equal(t, platform, info.Platform)
+	})
+
+	t.Run("arch", func(t *testing.T) {
+		t.Setenv("ARCH", arch)
+		info, err := nfpm.Parse(strings.NewReader("name: foo\narch: $ARCH"))
+		require.NoError(t, err)
+		require.Equal(t, arch, info.Arch)
+	})
 
 	t.Run("version", func(t *testing.T) {
 		os.Clearenv()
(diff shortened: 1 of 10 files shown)


djgilcrease (approved):
LGTM

nfpm_test.go
```diff
@@ -244,13 +244,31 @@ func TestOptionsFromEnvironment(t *testing.T) {
...
 		release         = "3"
 		version         = "1.0.0"
 		vendor          = "GoReleaser"
 		packager        = "nope"
 		maintainerEmail = "nope@example.com"
 	)
 
+	t.Run("platform", func(t *testing.T) {
+		os.Clearenv()
+		os.Setenv("OS", platform)
```
caarlos0, line 258: you can use `t.Setenv` instead

nfpm_test.go
```diff
@@ -244,13 +244,31 @@ func TestOptionsFromEnvironment(t *testing.T) {
...
+		info, err := nfpm.Parse(strings.NewReader("name: foo\nplatform: $OS"))
+		require.NoError(t, err)
+		require.Equal(t, platform, info.Platform)
+	})
+
+	t.Run("arch", func(t *testing.T) {
+		os.Clearenv()
+		os.Setenv("ARCH", arch)
```
caarlos0, line 266: same here

nfpm_test.go
afbjorklund: to be honest, I just copied from below

afbjorklund:
The tests seemed to be littering $TMPDIR, but maybe that's a feature ?

One can go inspect the *.rpm and *.deb files afterwards, if desired...

nfpm_test.go
caarlos0: oh, I'll change that then :D 

thansk

caarlos0:
> The tests seemed to be littering $TMPDIR, but maybe that's a feature ?

we should probably change to use `t.TempDir()` instead, I'll do it.

Thanks!

PR #603, opened 2023-01-15
fix(deb): check for empty provides value

Arsen6331:
This PR adds a check to the deb packager that returns an error if the `Provides` field contains an empty string. This check is needed because `dpkg` cannot parse provides lines with empty strings, and returns the following error when trying to install such a package:

```
E: Problem parsing Provides line
E: Error occurred while processing espanso-bin (NewVersion2)
E: Problem with MergeList /home/arsen/.cache/lure/pkgs/espanso-bin/espanso-bin_2.1.8-1_amd64.deb
E: The package lists or status file could not be parsed or opened.
```

This doesn't appear to be an issue with other packagers, so I've only added it for deb.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -89,14 +89,23 @@ func (*Deb) ConventionalExtension() string {
 	return ".deb"
 }
 
-// ErrInvalidSignatureType happens if the signature type of a deb is not one of
-// origin, maint or archive.
-var ErrInvalidSignatureType = errors.New("invalid signature type")
+var (
+	// ErrInvalidSignatureType happens if the signature type of a deb is not one of
+	// origin, maint or archive.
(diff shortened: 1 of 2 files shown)


caarlos0:
maybe its better to filter out empty provides instead?

Arsen6331:
Debian is fine when there are no provides, it only returns an error if there are provides and some of them are empty.

Arsen6331:
Or are you saying they should just be removed instead?

caarlos0:
I mean something like this: https://github.com/goreleaser/nfpm/compare/non-empty?expand=1

Arsen6331:
Yes, that seems better, thanks. I'll close this PR now.

PR #618, opened 2023-02-07
feat: Refactor contents handling and add tree content type.

### LINKED ISSUE #432: refactor deb and apk create tree
check https://github.com/goreleaser/nfpm/pull/431#issuecomment-1002129194 for details

erikgeiser:
This is quite a large PR that refactors the handling of `files.Contents` and introduces the new content type `tree` that copies a whole directory structure into the new package. It closes #432.

* **Why do `files.Contents` need to be handled differently?** Currently we have to carefully handle the contents in each packager in order to create implicit and explicit directories and files in the correct order and we have to explicitly ignore content types that are specific to other packagers. This PR moves this logic into a central place (the `files` package).
* **Why do you introduce the `tree` type in the same PR?** The tree behavior was already kind of implemented for `archlinux` packages so I had to already include this behavior in the refactor setup.
* **How does `tree` differ from a `file` where `src` is a directory?** When `src` is a directory all files in it or in any subfolder will be copied into the single directory specified in `dst`. In contrast, `tree` replicates the whole directory structure. As far as I can tell symlinks are also resolved in the case of `file` with a `src` dir while they are added as symlinks in the case of the `tree` type.

Let me give you an overview of the changes so you don't get lost in the giant diff:

* **files.PrepareForPackager:** This is the core of the refactor and it replaces and deprecates `ExpandContentGlobs`. The doc string describes what it does:
   - It filters out content that is irrelevant for the specified packager
   - It expands globs (if enabled) and file trees
   - It adds implicit directories (parent directories of files)
   - It adds ownership and other file information if not specified directly
   - It normalizes content source paths to be unix style paths
   - It normalizes content destination paths to be absolute paths with a trailing
     slash if the entry is a directory
  => As a result, each packager get only the contents that are relevant and **all** the contents the packager has to add including implicit directories
* `files` now declares and documents the content types in order to reduce string literals.
* `nfpm.Validate` is not called anymore but `nfpm.PrepareForPackager` is called instead. I changed `nfpm.Validate` so it does not modify the contents but still performs the checks so the behavior for users of the library does not break.


djgilcrease (approved):
LGTM

deb/deb.go
```diff
@@ -488,8 +439,21 @@ func copyToTarAndDigest(file *files.Content, tw *tar.Writer, md5w io.Writer) (in
 	return file.Size(), nil
 }
 
-func createChangelogInsideDataTar(tarw *tar.Writer, md5w io.Writer,
-	created map[string]bool, info *nfpm.Info,
+func withChangelogIfRequested(info *nfpm.Info) *nfpm.Info {
+	if info.Changelog != "" {
```
caarlos0, line 443: nitpick but could change the if to == "" and return early to lower the ident scope of the rest of the code

deb/deb.go
```diff
@@ -576,14 +533,14 @@ func createControl(instSize int64, md5sums []byte, info *nfpm.Info) (controlTarG
 	}
 
 	filesToCreate := map[string][]byte{
-		"control":   body.Bytes(),
-		"md5sums":   md5sums,
-		"conffiles": conffiles(info),
+		"./control":   body.Bytes(),
+		"./md5sums":   md5sums,
+		"./conffiles": conffiles(info),
```
caarlos0, line 538: I don't quite understand why this change is needed... could you add some details?

files/files.go
```diff
@@ -140,35 +187,106 @@ func (c *Content) Sys() interface{} {
...
+
+	return fmt.Sprintf("Content(%s)", strings.Join(properties, ","))
+}
+
+// ExpandContentGlobs gathers all of the real files to be copied into the
+// package.
+//
+// Deprecated: ExpandContentGlobs is deprecated in favor of PrepareForPackager
+// which also expands implicit directories and takes the packager into account.
 func ExpandContentGlobs(contents Contents, disableGlobbing bool) (files Contents, err error) {
```
caarlos0, line 228: does nfpm still uses this somewhere? if not, I think we can remove it... if anyone is using this, they likely shouldn't. 

on the same note, for a future PR/release, we might move this into the `internal` package.

caarlos0 (approved):
first of all, sorry for the delay in reviewing this... got carried away in life events here.

that said: this is awesome, thanks so much for the hard work on it!

I made a couple of very minor remarks, but it looks great to me!

deb/deb.go
erikgeiser: Wow, I'm normally super picky about always returning early, I can't belief I missed this one.

files/files.go
erikgeiser: How serious are we about a stable library API? If you want I can remove it, then we can also get rid of all the tests for `ExpandContentGlobs`.

caarlos0: I think we can remove it yes...

caarlos0:
Thanks @erikgeiser !

PR #630, opened 2023-03-04
feat: wrap pgp errors

caarlos0:
wrap pgp errors with `fmt.Errorf` so we can use `errors.Is`, and `require.ErrorIs`


nfpm_test.go
```diff
@@ -133,7 +133,7 @@ func TestPrepareForPackager(t *testing.T) {
 		asdFile := info.Overridables.Contents[0]
 		require.Equal(t, "/asd", asdFile.Destination)
 		require.Equal(t, files.TypeFile, asdFile.Type)
-		require.Equal(t, "-rw-r--r--", asdFile.FileInfo.Mode.String())
```
caarlos0, line 136: @erikgeiser I haven't debugged yet why this was failing on my machine, still need to take a look

djgilcrease (approved):
LGTM

Issue #638, opened 2023-03-24
archlinux package produces warnings when installed


j-tai:
### What happened?

When nfpm generates an archlinux package, and that package is installed, several warnings are printed by pacman:

```
loading packages...
resolving dependencies...
looking for conflicting packages...

Packages (1) foo-1.2.3-1

Total Installed Size:  0.00 MiB

:: Proceed with installation? [Y/n]
(1/1) checking keys in keyring                               [################################] 100%
(1/1) checking package integrity                             [################################] 100%
(1/1) loading package files                                  [################################] 100%
(1/1) checking for file conflicts                            [################################] 100%
(1/1) checking available disk space                          [################################] 100%
:: Processing package changes...
(1/1) installing foo                                         [################################] 100%
warning: file not found in file list for package foo. skipping extraction of usr/
warning: file not found in file list for package foo. skipping extraction of usr/share/
warning: file not found in file list for package foo. skipping extraction of usr/share/foo/
```

I expected the package to install without these warnings.

### How can we reproduce this?

On a machine running Arch Linux:
```console
$ git clone https://git.jtai.ca/repro/nfpm-2
$ cd nfpm-2
$ nfpm package -p archlinux
$ sudo pacman -U foo-1.2.3-1-any.pkg.tar.zst 
```

For reference, here is the nfpm.yaml:
```yaml
name: foo
arch: all
version: 1.2.3
maintainer: asdf

contents:
  - src: test.txt
    dst: /usr/share/foo/test.txt
```

### nfpm version

```bash
nfpm version nfpm version 2.27.1
commit: d30c22f50d30d92a8039bd14d789ad85d0019fd8
built at: 2023-03-20T01:43:40Z
built by: goreleaser
module version: v2.27.1, checksum: h1:+q7/aY+SCQjB5OKiqzdBVoFIhG1XnxRBU2Pxqt8qm3U=
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

j-tai:
It might be because the parent directories need to be added to the package's `.MTREE`.

caarlos0:
investigating this a bit...
<img width="1035" alt="CleanShot 2023-04-05 at 09 41 55@2x" src="https://user-images.githubusercontent.com/245435/230083492-ba677d24-55d0-47bb-b0c8-017f16125e89.png">

[...]

PR #645, opened 2023-04-05
fix(archlinux): .MTREE should have parent dirs as well

caarlos0:
this walks up the destination and creates the `.MTREE` entries for the parents too, otherwise packages are not installable by `pacman`.

if anyone knows a better way to do this, I'm all ears... 

closes https://github.com/goreleaser/nfpm/issues/638

### DIFF
--- a/arch/arch.go
+++ b/arch/arch.go
@@ -453,7 +453,7 @@ type MtreeEntry struct {
 
 func (me *MtreeEntry) WriteTo(w io.Writer) (int64, error) {
 	switch me.Type {
-	case files.TypeDir:
+	case files.TypeDir, files.TypeImplicitDir:
 		n, err := fmt.Fprintf(
 			w,
 			"./%s time=%d.0 mode=%o type=dir\n",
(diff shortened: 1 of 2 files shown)


erikgeiser:
I can’t really review this PR right now but from a first glance I think this should be fixed in our files library. There we create implicit directories (a dedicated content type) that we add for debs but not for rpms. So arch packages should work like debs in this regard.

erikgeiser:
We probably just need to add implicit dirs in the content type switch in WriteTo.

erikgeiser:
This should do the trick:
```go
 func (me *MtreeEntry) WriteTo(w io.Writer) (int64, error) {
        switch me.Type {
-   case files.TypeDir:
+ case files.TypeDir, files.TypeImplicitDir:
                n, err := fmt.Fprintf(
                        w,
                        "./%s time=%d.0 mode=%o type=dir\n",
```

However, the test does take the `info.Contents` into account. In my opinion, the `TestArchMtree` test should not check for implicit directories. We should add another test, that builds the entire package and then checks if the mtree contains implicit directories.

I'm sorry, but I don't have the time to rewrite the test, otherwise I would have already pushed the diff above.

caarlos0:
> This should do the trick:
> 
> ```go
>  func (me *MtreeEntry) WriteTo(w io.Writer) (int64, error) {
>         switch me.Type {
> -   case files.TypeDir:
> + case files.TypeDir, files.TypeImplicitDir:
>                 n, err := fmt.Fprintf(
>                         w,
>                         "./%s time=%d.0 mode=%o type=dir\n",
> ```
> 
> However, the test does take the `info.Contents` into account. In my opinion, the `TestArchMtree` test should not check for implicit directories. We should add another test, that builds the entire package and then checks if the mtree contains implicit directories.
> 
> I'm sorry, but I don't have the time to rewrite the test, otherwise I would have already pushed the diff above.

ah way better indeed, will work on it, thanks!

caarlos0:
@erikgeiser fwiw that doesn't work, will look into why later... but the root folders are still missing there...

erikgeiser:
Yeah I missed this one:
https://github.com/goreleaser/nfpm/blob/571a32b031d06d03fb7cbfab007f2506576c3f08/arch/arch.go#L177
The mtree entry should always be added.

caarlos0:
Ah right, I'm 100% blind 😂

caarlos0:
ight, all fixed, thanks for the help @erikgeiser

PR #668, opened 2023-06-02
feat: completion improvements

scop:
See individual commits for details.


scop:
CI fail is unrelated, fix in https://github.com/goreleaser/nfpm/pull/669

PR #675, opened 2023-06-07
feat: expand env variables in homepage field

scop:
For example, GitLab CI has `CI_PROJECT_URL` which can be useful here.


scop:
Aside, I'm beginning to wonder why not expand env vars in all fields :thinking:

djgilcrease:
> It might be easier to pass the config file through envsubst before passing it to nfpm thinking

Or at this point switch to all fields except contents being expanded.

PR #678, opened 2023-06-21
test: signatures on centos 7 and 8

caarlos0:
I wanted to see how far back nfpm still works... seems like centos 8 is the last one.

Now, on a subsequent PR, I'll try to upgrade openpgp lib... I'm fairly certain at least centos8 will break.

We can then decide next steps.


djgilcrease (approved):
LGTM

PR #680, opened 2023-06-23
feat: update protonmail/crypto

caarlos0:
As expected, this breaks signing for centos 8.
Not expected: it also broke it for centos 9.

Fedora seems to still work though.

Not sure what, if anything, we can do about this? Should we revert back to the main go/crypto package which supposedly still works?


djgilcrease:
I am unsure why this would break contos, we are using sha256 for signing nor sha1 which GopenPGP still supports

caarlos0:
> I am unsure why this would break contos, we are using sha256 for signing nor sha1 which GopenPGP still supports

as far as I gathered, centos still uses an old version of the signature, which is supported by golang/x/crypto/openpgp, but not by the protonmail fork (because it is very old)

https://github.com/sassoftware/go-rpmutils/pull/21#issuecomment-1100266607

djgilcrease:
> > I am unsure why this would break contos, we are using sha256 for signing nor sha1 which GopenPGP still supports
> 
> as far as I gathered, centos still uses an old version of the signature, which is supported by golang/x/crypto/openpgp, but not by the protonmail fork (because it is very old)
> 
> [sassoftware/go-rpmutils#21 (comment)](https://github.com/sassoftware/go-rpmutils/pull/21#issuecomment-1100266607)

https://github.com/ProtonMail/go-crypto/issues/164 maybe try `replace github.com/ProtonMail/go-crypto => github.com/pgpkeys-eu/go-crypto v0.0.0-20230506215654-16de0cc09494`

caarlos0:
hmmm still getting 


```
#11 0.562 /tmp/foo.rpm: digests SIGNATURES NOT OK
```

so maybe there is something else wrong

caarlos0:
Okay, for reference:


```
❯ docker run -it --rm quay.io/centos/centos:stream9-minimal rpm --version
RPM version 4.16.1.3

❯ docker run -it --rm quay.io/centos/centos:stream8 rpm --version
RPM version 4.14.3

❯ docker run -it --rm fedora rpm --version
RPM version 4.18.1

```


RPM 4.18 seems to work, the older versions don't.

caarlos0:
4.18 changelog: https://rpm.org/wiki/Releases/4.18.0

there's a lot of signature and openpgp related changes and fixes.


I'm leaning towards supporting RPM 4.18+ for signed packages for now on, and that's it 🤔

caarlos0:
so it seems rpm 4.17+ is OK 🤔

caarlos0:
okay, so, porting back to the deprecated crypto package works, as long as you don't need to set the signing key id, because apparently there is no way of setting it in that lib


soo... I made a compromise:

- if the user set the signing key id, it'll use protonmail's fork and resulting packages will not work on old rpms
- if the user don't set it, it uses the deprecated version, which works

which I think is pretty confusing... and not sure if whether or not is a good idea at all 


eager to hear what you all think @djgilcrease @erikgeiser

caarlos0:
PS: the very same tests pass in my machine 🤔

caarlos0:
interesting... https://github.com/rpm-software-management/rpm/issues/2351


so, basically, the old package breaks new RPMs, the new version breaks old RPMs... so is just a matter of choosing which ones to break?

:pain:

caarlos0:
another alternative (that honestly I almost think is better) is to copy the needed code from the versions that work to our codebase 🤔

caarlos0:
continuing my investigation:

- https://github.com/ProtonMail/go-crypto/commit/c05353c2d58c90ebc6e81dfabef7e3ba48137059: passes on 4.14+
- https://github.com/ProtonMail/go-crypto/commit/cc34b1f6908b80f7c0d6d0ca0065c08ccc6b83f3: passes on 4.18+
- https://github.com/ProtonMail/go-crypto/commit/317b2172f8a550594a08017ca5366acf9a9fa5e3: passes on 4.18+
- https://github.com/ProtonMail/go-crypto/commit/5ea50bddae69644782cb279f37b3ab27ce6f4689: passes on 4.18+
- https://github.com/ProtonMail/go-crypto/commit/428f8eabeeb3a9b3fe1d2485cc516efeddcc3358: passes on 4.18+
- https://github.com/ProtonMail/go-crypto/commit/a4f6767435ab31ee45e3097fc96cae080824b7e0: passes on 4.17+
- https://github.com/ProtonMail/go-crypto/commit/f2de896918a33ef92f4b50fafaf4172e9f9d2896: passes on 4.17+
- https://github.com/ProtonMail/go-crypto/commit/3fbb1f12458ced1e64a5e4027e340f2f7b8d2a85: (latest) passes on 4.17+

So, I think we can presume that's something related to the critical bit of the
signatures, right?

caarlos0:
okay, this will do the trick: https://github.com/ProtonMail/go-crypto/pull/175

PR #681, opened 2023-06-23
refactor: using golang.org/x/crypto for signing rpms

caarlos0:
see #680

with this, I think we would be able to keep things up to date more easily.

afaik, the problem is that protonmail's crypto packages do not support some old version of the signatures, which centos still uses...

so, we either drop support for centos, or use golang.org/x/crypto for signing rpms.


djgilcrease:
The problem with golang.org/x/crypto is the openpgp is [deprecated](https://github.com/golang/crypto/commit/0ba0e8f031222feafe2c6fe04ba9c3da11dd361a) and it seems centos back to 7 supports (sha256)[https://centosfaq.org/centos/signing-rpm-packages-with-sha256/] we may just need to update (rpmpack)[https://github.com/goreleaser/nfpm/tree/main/internal/rpmpack] to understand `%_gpg_digest_algo sha256`

erikgeiser:
Given our limited use of PGP, I don't think the deprecation is that much of an issue, as security fixes are still provided. That said, I also don't mind dropping support for the old Centos versions, especially because I think security backports for distros that are that old are a farce anyway. If we can solve it by improving `rpmpack`, that would also be nice.

djgilcrease:
Also rpm supports [sha512](https://github.com/rpm-software-management/rpm/blob/92eadae94c48928bca90693ad63c46ceda37d81f/rpmio/rpmpgp.h#L267C5-L267C23) but we are [hardcoding it to sha256](https://github.com/goreleaser/nfpm/blob/main/internal/rpmpack/tags.go#L29)

PR #684, opened 2023-06-27
fix: version output

figsoda:
`nfpm version nfpm version <version>` -> `nfpm version <version>`


caarlos0:
oh, good catch! thanks!

PR #693, opened 2023-07-14
docs: improve deb fields var expansion example

scop:
`$CI_REPOSITORY_URL` is not quite appropriate for `Vcs-Git`, use `$CI_PROJECT_URL` with `Vcs-Browser` instead.

- https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
- https://www.debian.org/doc/debian-policy/ch-controlfields.html#version-control-system-vcs-fields


caarlos0:
nice, thank you!

PR #698, opened 2023-08-12
feat(files): accept `doc`, `licen[cs]e`, and `readme` with non-rpm

scop:
See commentary in code for elaboration.

This PR might be a bit incomplete: untested, no test coverage I think, and perhaps `RPM` should be dropped from the type names (with aliases for backwards compatibility) to make it clearer that files with them do get included with non-rpm packagers too.

Anyway, submitting here already in this form for discussion, before spending more time on it.

One example case is at https://github.com/aakso/ssh-inscribe/blob/7922b3e15f5fd9569c927f1464982c6f9e870f5b/.goreleaser.yaml#L69-L82. Note how the `LICENSE` and `README.md` entries need to be duplicated for deb and rpm; deb without the type info, rpm with it. It would get worse if there were more packager types enabled.

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -37,13 +37,13 @@ const (
 	TypeConfigNoReplace = "config|noreplace"
 	// TypeGhost is the type of an RPM ghost file which is ignored by other packagers.
 	TypeRPMGhost = "ghost"
-	// TypeRPMDoc is the type of an RPM doc file which is ignored by other packagers.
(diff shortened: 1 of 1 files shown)


caarlos0:
hmm, yes, I think that if we do this, we need to type-alias et al to make it compatible with previous versions, or bump nfpm to v3

PR #721, opened 2023-10-04
feat: env expand contents

### LINKED ISSUE #719: Support env var expansion for `contents.src`

I have a directory structure produced by Docker/BuildKit multi-platform build:
```
tmp/build/
├── linux_amd64
│   └── my-bin
└── linux_arm64
    └── my-bin
```

I want to be able to use the single `nfpm.yml` configuration file for both architectures. Unfortunately, it is not possible to use environment variable expansion in `contents.src`.

```yaml
arch: ${ARCH}
contents:
  - src: tmp/build/linux_${ARCH}/my-bin
    dst: /usr/bin/my-bin
```

Using multiple `nfpm.yml` files. Using external templates. I would like to avoid both those alternatives.

### Search

### Code of Conduct

### Additional context

I should be able to contribute that change if that's something you want to have.

caarlos0:
closes #719

### DIFF
--- a/www/docs/configuration.md
+++ b/www/docs/configuration.md
@@ -253,6 +253,12 @@ contents:
     file_info:
       mode: 0700
 
+  # Using `expand: true`, environment variables will be expanded in both
+  # src and dst.
+  - dst: /usr/local/bin/${NAME}
+    src: "${NAME}"
+    expand: true
+
 # Umask to be used on files without explicit mode set.
 #
 # By default, nFPM will inherit the mode of the original file that's being


djgilcrease:
Careful with this as we have rejected this in the past because `$` is a valid path character. So we need tests for that and need to document that if a `$` is in the path it needs to be escaped.

AlekSi (approved):
:shipit:

djgilcrease:
> Another option is to opt-in into expanding:
> 
I like this solution, that way it is explicit that the user knows it will expand env vars

caarlos0:
pushed

djgilcrease (approved):
LGTM

PR #733, opened 2023-11-06
Correct GOARCH docs

cognifloyd:
The docs had an extra `v` in the GOARCH+GOARM values.

Drop the `v` to match the actual translation tables:

https://github.com/goreleaser/nfpm/blob/6c241370ccdd5718730556f2b8c13e4733d0fda7/apk/apk.go#L65-L66 https://github.com/goreleaser/nfpm/blob/6c241370ccdd5718730556f2b8c13e4733d0fda7/arch/arch.go#L46-L48 https://github.com/goreleaser/nfpm/blob/6c241370ccdd5718730556f2b8c13e4733d0fda7/deb/deb.go#L40-L42 https://github.com/goreleaser/nfpm/blob/6c241370ccdd5718730556f2b8c13e4733d0fda7/rpm/rpm.go#L63-L65

### DIFF
--- a/www/docs/goarch-to-pkg.md
+++ b/www/docs/goarch-to-pkg.md
@@ -21,9 +21,9 @@ Thank you!
 | `386` | `i386` |
 | `amd64` | `x86_64` |
 | `arm64` | `arm64` |
-| `armv5` | `armel` |
-| `armv6` | `armhf` |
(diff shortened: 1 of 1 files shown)


caarlos0:
good catch, thank you!

PR #742, opened 2023-11-24
fix(deb): blank line in package description

abemedia:
Blank lines in the deb's control file description are valid but need to be encoded with a space followed by a dot (`.`).

See https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-description.
> Those containing a single space followed by a single full stop character. These are rendered as blank lines. This is the only way to get a blank line.

I used nfpm to package one of my apps and it was broken as it did not add the dots onto the blank lines. This PR fixes that.

### DIFF
--- a/deb/deb_test.go
+++ b/deb/deb_test.go
@@ -523,7 +523,7 @@ func TestMultilineFields(t *testing.T) {
 		Info: nfpm.WithDefaults(&nfpm.Info{
 			Name:        "multiline",
 			Arch:        "riscv64",
-			Description: "This field is a\nmultiline field\nthat should work.",
+			Description: "This field is a\nmultiline field\n\nthat should work.",
 			Priority:    "extra",
 			Version:     "1.0.0",
 			Section:     "default",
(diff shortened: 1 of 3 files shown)


deb/deb_test.go
```diff
@@ -523,7 +523,7 @@ func TestMultilineFields(t *testing.T) {
 		Info: nfpm.WithDefaults(&nfpm.Info{
 			Name:        "multiline",
 			Arch:        "riscv64",
-			Description: "This field is a\nmultiline field\nthat should work.",
+			Description: "This field is a\nmultiline field\n\nthat should work.",
```
caarlos0, line 526: shouldn't it be?
```golang
			Description: "This field is a\nmultiline field\n.\nthat should work.",
```

abemedia: This PR adds the dot when creating the control file. After all, you wouldn't want the dot for RPM etc.

caarlos0: Ah right, I'm blind

caarlos0:
Thank you 🙏

PR #747, opened 2023-12-05
fix: Ensure reproducable Deb package

keliramu:
Each time building exaclty from the same source,
ensure exaclty the same size deb package is generated.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -391,14 +391,14 @@ func createFilesInsideDataTar(info *nfpm.Info, tw *tar.Writer) (md5buf bytes.Buf
 				Format:   tar.FormatGNU,
 				Uname:    file.FileInfo.Owner,
 				Gname:    file.FileInfo.Group,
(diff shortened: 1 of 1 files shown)


deb/deb.go
```diff
@@ -554,18 +554,27 @@ func createControl(instSize int64, md5sums []byte, info *nfpm.Info) (controlTarG
 		return nil, err
 	}
 
-	filesToCreate := map[string][]byte{
-		"./control":   body.Bytes(),
-		"./md5sums":   md5sums,
-		"./conffiles": conffiles(info),
+	// ensure predefined sort order of these items
+	filesToCreateNames := []string{
```
keliramu, line 558: Go `map[string]` does not ensure predefined sort order when iterated via `for ... range ...` and that is the cause of `control.tar.gz` is different size when executed on the same input.

caarlos0:
gonna merge this and continue working on it, thanks for the PR

PR #748, opened 2023-12-06
feat: reproducible packages

### LINKED ISSUE #734: Add support for SOURCE_DATE_EPOCH

If a package is re-built with the same contents, but at two different times, the two packages will have binary differences because of the embedded timestamps.

I need to be able to build packages in a way that is completely reproducible, so that the same inputs gives me byte-for-byte the same package output.

For the files I'm packaging, I can set the `mtime` via the config file, or by changing the mtime on the file tree before running nfpm. However, there is no current mechanism for me to manage the timestamps generated within nFPM.

Make the package build process reproducible by using the standard `SOURCE_DATE_EPOCH` env var (if available) instead of using `time.Now()`.

More specifically, I think we can replace all of the `time.Now()` occurrences with one value calculated early on. Then, adjust that calculation to use `SOURCE_DATE_EPOCH` instead, if available.

This is the go example of calculating that:
https://reproducible-builds.org/docs/source-date-epoch/#go

```go
import (
        "fmt"
        "os"
        "strconv"
        "time"
)

[...]

source_date_epoch := os.Getenv("SOURCE_DATE_EPOCH")
var build_date string
if source_date_epoch == "" {
        build_date = time.Now().UTC().Format(http.TimeFormat)
} else {
        sde, err := strconv.ParseInt(source_date_epoch, 10, 64)
        if err != nil {
                panic(fmt.Sprintf("Invalid SOURCE_DATE_EPOCH: %s", err))
        }
        build_date = time.Unix(sde, 0).UTC().Format(http.TimeFormat)
}
```

We can probably also use this `build_date` for a default `mtime` of package contents, but that is NOT what this issue is about. This issue is about controlling the current usages of `time.Now()`. These are the current instances that github search shows:
https://github.com/search?q=repo%3Agoreleaser%2Fnfpm+time.now&type=code

A new cli flag, or any other non-standard option means any automation has to learn a new trick instead of reusing the standard env var. More detailed description on why the env var is better: https://wiki.debian.org/ReproducibleBuilds/StandardEnvironmentVariables#A.22We.27ll_add_a_command-line_flag_instead.22

### Search

### Code of Conduct

### Additional context

More background on why reproducible builds are important: https://reproducible-builds.org/

The spec for `SOURCE_DATE_EPOCH`: https://reproducible-builds.org/specs/source-date-epoch/

Apparently, `dpkg` automatically sets `SOURCE_DATE_EPOCH` when building packages based on the latest entry in the `debian/changelog` file. Many other tools have adopted that standard as well:
https://reproducible-builds.org/docs/source-date-epoch/#reading-the-variable

### LINKED ISSUE #744: `nfpm` does not ensure reproducible builds
### What happened?

Implemented sample `nfpm` template to create `deb` package (as a bash script to create deb package).
Executed above mentioned script in a loop of 9 iterations: create package, check size, remove package.
Getting stable results: at least 2 times out of 9 `deb` size differs from previous iteration.
```
~/nfpm-consistent-results-not$ ./test-deb-size.sh 
Create DEB multiple times to check if resulting size stays the same
Create DEB in loop: 1..9
DEB i: 1; now DEB size: 862
DEB i: 2; now DEB size: 864 ::: GOT different size!!! prev: [862]
DEB i: 3; now DEB size: 862 ::: GOT different size!!! prev: [864]
DEB i: 4; now DEB size: 864 ::: GOT different size!!! prev: [862]
DEB i: 5; now DEB size: 864
DEB i: 6; now DEB size: 864
DEB i: 7; now DEB size: 864
DEB i: 8; now DEB size: 864
DEB i: 9; now DEB size: 864
Done. Got diffs: 3
```

Here is the public `github` repo with scripts to reproduce this problem: https://github.com/keliramu/nfpm-consistent-results-not

### nfpm version

```bash
nfpm --version
       _____ ____  __  __
 _ __ |  ___|  _ \|  \/  |
| '_ \| |_  | |_) | |\/| |
| | | |  _| |  __/| |  | |
|_| |_|_|   |_|   |_|  |_|
nfpm: a simple and 0-dependencies deb, rpm, apk and arch linux packager written in Go
https://nfpm.goreleaser.com

GitVersion:    2.34.0
GitCommit:     6a16fcd38aa41ec2a5fea50728d22258ab9e2580
GitTreeState:  false
BuildDate:     2023-10-23T11:26:52Z
BuiltBy:       goreleaser
GoVersion:     go1.21.3
Compiler:      gc
ModuleSum:     h1:fKoHucBOcmW2CkIDj3gZZ4grJGRRoed7eRzAztWa3xo=
Platform:      linux/amd64
```

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
This adds a new `mtime` fields to the root of the yaml configuration file, and replace all occurrences of `time.Now()` with its value.

This should help making packaging reproducible.

Also, if that value is empty, it will default to `$SOURCE_DATE_EPOCH` as per reproducible-builds.org

It also fixes several instances in which ordering could change, changing the contents of the package slightly (namely, usage of maps).

closes #744
closes #734

### DIFF
--- a/arch/arch.go
+++ b/arch/arch.go
@@ -16,9 +16,9 @@ import (
 
 	"github.com/goreleaser/nfpm/v2"
 	"github.com/goreleaser/nfpm/v2/files"
+	"github.com/goreleaser/nfpm/v2/internal/maps"
 	"github.com/klauspost/compress/zstd"
 	"github.com/klauspost/pgzip"
-	"golang.org/x/exp/maps"
 )
 
 var ErrInvalidPkgName = errors.New("archlinux: package names may only contain alphanumeric characters or one of ., _, +, or -, and may not start with hyphen or dot")
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -22,10 +22,10 @@ import (
 	"github.com/goreleaser/nfpm/v2"
 	"github.com/goreleaser/nfpm/v2/deprecation"
 	"github.com/goreleaser/nfpm/v2/files"
+	"github.com/goreleaser/nfpm/v2/internal/maps"
 	"github.com/goreleaser/nfpm/v2/internal/sign"
 	"github.com/klauspost/compress/zstd"
 	"github.com/ulikunitz/xz"
-	"golang.org/x/exp/maps"
 )
 
 const packagerName = "deb"
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/stretchr/testify v1.8.4
 	github.com/ulikunitz/xz v0.5.11
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8
+	golang.org/x/exp v0.0.0-20231206192017-f3f8817b8deb
 	gopkg.in/yaml.v3 v3.0.1
 )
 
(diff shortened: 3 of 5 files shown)


caarlos0:
PS: jury not set on the field name, feel free to suggest others

cognifloyd (approved):
OOH! Thank you! This looks fantastic.

djgilcrease (approved):

cognifloyd (approved):

PR #788, opened 2024-02-21
feat(rpm): add support for verify scriptlet

nickajacks1:
RPM allows users to add custom verify scriptlets that are run by `rpm -V` in addition to the normal checksum validation. This is useful when there are additional factors that indicate a valid installation, like, for example, checking for valid contents of configuration files.


caarlos0:
Thanks!

Issue #801, opened 2024-03-18
s390x support


Repana-Chowdappa:
### Is your feature request related to a problem? Please describe.

s390x artifacts are currently not being published to the releases page.

### Describe the solution you'd like

We would like to open a PR for ppc64le support.

### Describe alternatives you've considered

s390x seems to be there already in some source files.

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

Hi team,
I am from the IBM Z (s390x) porting team. We would like to open a PR for s390x support so that the artifacts are published to the releases page for subsequent releases. I searched through the source files but could not find where to make this change. Can you please help?

caarlos0:
someone from ibm opened the exact same issue for ppc64 a couple of months ago...

Repana-Chowdappa:
@caarlos0 Yes, I saw the power(ppc64) pull request, I see there is only 1 change in` .goreleaser.yml ` and you have completed it as part of https://github.com/goreleaser/nfpm/commit/37b28fd4a368a94e4e285bdabbd3c1641f28cff4

Repana-Chowdappa:
Please do let me know if this single change work for s390x support?

Repana-Chowdappa:
@caarlos0 - When we will be able to get the s390x binary on https://github.com/goreleaser/nfpm/releases/download ?

PR #802, opened 2024-03-19
add s390x support

lysliu:
Fix: #801

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,6 +24,7 @@ builds:
       - amd64
       - arm64
       - ppc64le
+      - s390
     mod_timestamp: "{{ .CommitTimestamp }}"
     flags:
       - -trimpath


.github/workflows/build.yml
```diff
@@ -69,6 +69,7 @@ jobs:
     env:
       DOCKER_CLI_EXPERIMENTAL: "enabled"
       NO_TEST_PPC64LE: "true"
+      NO_TEST_S390X: "true"
```
caarlos0, line 72: I don't think this is needed... we did that for ppc64le because some tests were failing only inside docker containers or something in that sense...

for s390x, it seems that we are already running the tests (line 67), so they are in fact passing already.

Repana-Chowdappa:
@lysliu @caarlos0 Can we specify the arch value as "s390x" instead of s390?

caarlos0:
done in 507ad79260d3c3c851c7f0340d3782240846de91

Issue #799, opened 2024-03-16
[apk] Cannot sign package


cnt0:
### What happened?

Hello. I'm trying to sign an apk package with key generated by `abuild-keygen`, which is a standard tool for alpine linux distribution.

It fails with the following message:
```
signing error: parse PKCS1 private key: x509: failed to parse private key (use ParsePKCS8PrivateKey instead for this key format)
```

### How can we reproduce this?

1. generate a keypair using the following command: `$ abuild-keygen -a -i`, it'll be saved to `~/.abuild` and `/etc/apk/keys`
2. add signature information to `nfpm.yaml` (the exact key name will indeed be different):
```yaml
apk:
  signature:
    key_file: /home/user/.abuild/user@email.com-xxxxxxxx.rsa
    key_name: user@email.com-xxxxxxxx
```
3. try generating a package:
```bash
$ nfpm pkg --packager apk --target package.apk
```

expected result: apk successfully generated

actual result: the aforementioned message
```
signing error: parse PKCS1 private key: x509: failed to parse private key (use ParsePKCS8PrivateKey instead for this key format)
```

this issue can be solved by manually converting the key to required PKCS1 format: `openssl -in key.pkcs8 -out key.pkcs1 -traditional`, but still, I believe the default can be changed.

### nfpm version

```bash
GitVersion:    2.35.3
GitCommit:     unknown
GitTreeState:  unknown
BuildDate:     unknown
BuiltBy:       unknown
GoVersion:     go1.22.1
Compiler:      gc
ModuleSum:     unknown
Platform:      linux/amd64
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

openssl version, just in case:
```sh
$ openssl version
OpenSSL 3.1.5 30 Jan 2024 (Library: OpenSSL 3.1.5 30 Jan 2024)
```
Also, I verified that the key generated by `abuild-keygen` is indeed a PEM-encoded RSA-4096 private key.

abuild-keygen version: `abuild-keygen 3.12.0-r4`

caarlos0:
good find!

if anyone wants to PR this, please feel free to :)

cnt0:
I'll make a PR in case of successful time allocation, which is non-deterministic :/

[...]

PR #804, opened 2024-03-26
feat: support PKCS8 keys for APK signing

cnt0:
Call correct parsing function for given RSA key depending on its PEM header. 

So we can use both PKCS1 and PKCS8 keys to sign APK files.

Fixes https://github.com/goreleaser/nfpm/issues/799


caarlos0:
amazing, thank you!

caarlos0:
Thanks!

PR #822, opened 2024-05-11
feat: Added "excludes" option to prevent packing unnecessary files

KenichiTanino:
For the type:tree option or when using glob settings, an excludes setting has been added.
If the excludes setting is specified, the target files will not be copied to the dst directory.

for examples:

```
contents:
  - src: ./service
    dst: /service
    type: tree
    excludes:
      - /service/folder
      - /service/etc
```


caarlos0:
sorry, not sure i want to complicate even more the files interface

KenichiTanino:
> sorry, not sure i want to complicate even more the files interface

Understood. Then I will close the issue.

PR #839, opened 2024-07-05
test: latest debian

caarlos0:
this fixes dpkg-sig tests, but breaks debsig

I have no idea why... anyone?


caarlos0:
fixed in 666ef77

PR #858, opened 2024-09-17
fix: undefined variable 'staging-pkgs'

robsliwi:
This commit fixes the undefined staging-pkgs like in the GitHub Actions runs:
[example](https://github.com/goreleaser/nfpm/actions/runs/10854480085/job/30124997498#step:4:87)


caarlos0:
thanks! my bad

PR #899, opened 2025-01-10
test: test centos10, fedora 40+

caarlos0:
- for the time being, commented out centos 9 uses RPM 4.16 which does not work correctly if key issuer id is critical (see https://github.com/ProtonMail/go-crypto/issues/263)
- added tests for centos 10
- removed EOL fedora 34, 36, 38
- added fedora 40, 41


caarlos0:
gonna merge this for now as we need to update go-crypto for a CVE :(

PR #948, opened 2025-07-08
feat: Add "force_implicit_dirs" to dir/tree.

volkadav:
This allows users to specify a list of directories to treat akin to the list of system directories, i.e. things that the constructed package should *not* take ownership of in e.g. the rpm database. (I'm 100% open to a better name for that, btw!)

Example use case:
1) user belongs to FooCorp
2) FooCorp has many packages to install under /opt/foocorp
   e.g. /opt/foocorp/service[A,B,C...]
3) Using tree w/ forced implicit dirs keeps generated package(s)
   from claiming ownership of /opt/foocorp so that the
   service[A,B,C...] packages can be bundled and installed
   separately without path conflicts.

The particular scenario I ran into this for was with RHEL8+ being much pickier about each dir being owned by one and only one package. (We pair this with having one "foocorp-dirs" package that just owns the shared dir creation, that's used as a pre-req for the other packages.) We could have done something like explicitly list every sub-dir needed or every file for all the service packages, but using `tree` was _much_ more concise and flexible with respect to package contents changing over time, and in effect this is giving the user the ability to specify dirs to treat like the hard-coded list of system directories already checked towards the same end.

I have run `task ci` and the normal tests pass. There seems to be some issue with `opkg` going away on the latest OpenWRT images that keeps the `ipk` acceptance test from working though:
```
#9 ERROR: process "/bin/sh -c opkg install /tmp/foo.ipk" did not complete successfully: exit code: 127
                                ------
                                 > [min 1/1] RUN opkg install /tmp/foo.ipk:
                                0.287 /bin/sh: opkg: not found
                                ------
                                ipk.dockerfile:12
                                --------------------
                                  10 |     # ---- minimal test ----
                                  11 |     FROM test_base AS min
                                  12 | >>> RUN opkg install /tmp/foo.ipk
                                  13 |
                                  14 |
                                --------------------
```
Sadly I don't know enough about OpenWRT/opkg/Docker to guess what the right solution would be for that, so I haven't included any changes for it.

Thank  you for a cool tool! :)

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -68,12 +68,13 @@ const (
 // Content describes the source and destination
 // of one file to copy into a package.
 type Content struct {
-	Source      string           `yaml:"src,omitempty" json:"src,omitempty"`
-	Destination string           `yaml:"dst" json:"dst"`
-	Type        string           `yaml:"type,omitempty" json:"type,omitempty" jsonschema:"enum=symlink,enum=ghost,enum=config,enum=config|noreplace,enum=config|missingok,enum=doc,enum=license,enum=licence,enum=readme,enum=dir,enum=tree,enum=config|tree,enum=config|noreplace|tree,enum=config|missingok|tree,enum=,default="`
-	Packager    string           `yaml:"packager,omitempty" json:"packager,omitempty"`
-	FileInfo    *ContentFileInfo `yaml:"file_info,omitempty" json:"file_info,omitempty"`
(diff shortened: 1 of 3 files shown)


erikgeiser (commented):
I like this proposal. However, we would need a test as well. Don't worry, I'll update the PR myself.

erikgeiser:
@caarlos0 is it okay if I merge this?

caarlos0:
@erikgeiser of course 🙏🏻

PR #961, opened 2025-08-21
Improve version extraction of `systemctl`

supcik:
On my Raspberry Pi, the first line of `systemctl --version` looks like this :

```
systemd 252 (252.36-1~deb12u1)
```

The `systemd_version` variable should be an integer because it is compared to 231 a couple of line further.

In the proposed change, I use `grep` instead of `sed`. The `grep` command prints all integers from the line, and the second `head -1` keeps the first integer only.


caarlos0:
Thanks!

PR #1000, opened 2025-11-20
The debian image now supports zstd compression

afbjorklund:
Old comment was from `debian:11`, supported since `debian:12`

Current test image is already based on `debian:13` (`trixie`).


caarlos0:
thank you!

Issue #628, opened 2023-03-01
archlinux package signing 


tonobo:
### Is your feature request related to a problem? Please describe.

Guess some refactoring is needed according to https://github.com/goreleaser/nfpm/pull/543#issuecomment-1245597192

### Describe the solution you'd like

Add signing

### Describe alternatives you've considered

No way ;)

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1065, opened 2026-04-03
feat: add support for signing archlinux packages

flecno:
Archlinux packages can now be signed with a detached PGP signature, producing a binary .sig file alongside the package — matching the format expected by pacman-key --verify.

The signing reads back the finalized .pkg.tar.zst from disk via info.Target to avoid buffering the entire package in memory. The passphrase is taken from $NFPM_ARCHLINUX_PASSPHRASE with a fallback to $NFPM_PASSPHRASE, consistent with deb/rpm/apk.

Also adds sign.PGPDetachedSignWithKeyID, a streaming variant of PGPSignerWithKeyID that accepts an io.Reader instead of []byte.

See #628

### DIFF
--- a/arch/arch.go
+++ b/arch/arch.go
@@ -3,6 +3,7 @@ package arch
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"crypto/md5"
 	"crypto/sha256"
@@ -19,6 +20,7 @@ import (
 	"github.com/goreleaser/nfpm/v2/files"
 	"github.com/goreleaser/nfpm/v2/internal/maps"
 	"github.com/goreleaser/nfpm/v2/internal/modtime"
+	"github.com/goreleaser/nfpm/v2/internal/sign"
 	"github.com/klauspost/compress/zstd"
 	"github.com/klauspost/pgzip"
 )
@@ -134,10 +136,8 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	defer zw.Close()
 
 	tw := tar.NewWriter(zw)
-	defer tw.Close()
 
 	entries, totalSize, err := createFilesInTar(info, tw)
 	if err != nil {
@@ -157,7 +157,32 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 		return fmt.Errorf("create mtree: %w", err)
 	}
 
-	return createScripts(info, tw)
+	err = createScripts(info, tw)
+	if err != nil {
+		return fmt.Errorf("create scripts: %w", err)
+	}
+
+	// finalize the tar/zstd writer before creating the signature
+	tw.Close()
+	zw.Close()
+
+	if info.ArchLinux.Signature.KeyFile != "" || info.ArchLinux.Signature.SignFn != nil {
+		sig, err := createSignature(info)
+		if err != nil {
+			return &nfpm.ErrSigningFailure{
+				Err: fmt.Errorf("create signature: %w", err),
+			}
+		}
+
+		err = createSignatureFile(info, sig)
+		if err != nil {
+			return &nfpm.ErrSigningFailure{
+				Err: fmt.Errorf("create signature file: %w", err),
+			}
+		}
+	}
+
+	return nil
 }
 
 // ConventionalExtension returns the file name conventionally used for Arch Linux packages
@@ -400,6 +425,42 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 		MD5:         md5Hash.Sum(nil),
 		SHA256:      sha256Hash.Sum(nil),
 	}, nil
+}
+
+func createSignature(info *nfpm.Info) ([]byte, error) {
+	f, err := os.Open(info.Target)
+	if err != nil {
+		return nil, fmt.Errorf("open package for signing: %w", err)
+	}
+	defer f.Close()
+
+	data := bufio.NewReader(f)
+
+	var sig []byte
+	if signFn := info.ArchLinux.Signature.SignFn; signFn != nil {
+		sig, err = signFn(data)
+	} else {
+		sig, err = sign.PGPDetachedSignWithKeyID(data, info.ArchLinux.Signature.KeyFile, info.ArchLinux.Signature.KeyPassphrase, info.ArchLinux.Signature.KeyID)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
+		}
+	}
+	return sig, nil
+}
+
+func createSignatureFile(info *nfpm.Info, sig []byte) error {
+	sigFile, err := os.Create(info.Target + ".sig")
+	if err != nil {
+		return fmt.Errorf("create signature file: %w", err)
+	}
+	defer sigFile.Close()
+
+	_, err = sigFile.Write(sig)
+	if err != nil {
+		return fmt.Errorf("write signature to file: %w", err)
+	}
+
+	return nil
 }
 
 func writeKVPairs(w io.Writer, pairs map[string]string) error {
--- a/arch/arch.go
+++ b/arch/arch.go
@@ -3,6 +3,7 @@ package arch
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"crypto/md5"
 	"crypto/sha256"
@@ -19,6 +20,7 @@ import (
 	"github.com/goreleaser/nfpm/v2/files"
 	"github.com/goreleaser/nfpm/v2/internal/maps"
 	"github.com/goreleaser/nfpm/v2/internal/modtime"
+	"github.com/goreleaser/nfpm/v2/internal/sign"
 	"github.com/klauspost/compress/zstd"
 	"github.com/klauspost/pgzip"
 )
@@ -158,7 +160,34 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 		return fmt.Errorf("create mtree: %w", err)
 	}
 
-	return createScripts(info, tw)
+	if err = createScripts(info, tw); err != nil {
+		return fmt.Errorf("create scripts: %w", err)
+	}
+
+	// finalize the tar/zstd writer before creating the signature
+	if err = tw.Close(); err != nil {
+		return fmt.Errorf("closing data tarball: %w", err)
+	}
+	if err = zw.Close(); err != nil {
+		return fmt.Errorf("closing zstd writer: %w", err)
+	}
+
+	if info.ArchLinux.Signature.KeyFile != "" || info.ArchLinux.Signature.SignFn != nil {
+		sig, err := createSignature(info)
+		if err != nil {
+			return &nfpm.ErrSigningFailure{
+				Err: fmt.Errorf("create signature: %w", err),
+			}
+		}
+
+		if err = createSignatureFile(info, sig); err != nil {
+			return &nfpm.ErrSigningFailure{
+				Err: fmt.Errorf("create signature file: %w", err),
+			}
+		}
+	}
+
+	return nil
 }
 
 // ConventionalExtension returns the file name conventionally used for Arch Linux packages
@@ -401,6 +430,45 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 		MD5:         md5Hash.Sum(nil),
 		SHA256:      sha256Hash.Sum(nil),
 	}, nil
+}
+
+func createSignature(info *nfpm.Info) ([]byte, error) {
+	f, err := os.Open(info.Target)
+	if err != nil {
+		return nil, fmt.Errorf("open package for signing: %w", err)
+	}
+	defer f.Close()
+
+	data := bufio.NewReader(f)
+
+	var sig []byte
+	if signFn := info.ArchLinux.Signature.SignFn; signFn != nil {
+		sig, err = signFn(data)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
+		}
+	} else {
+		sig, err = sign.PGPDetachedSignWithKeyID(data, info.ArchLinux.Signature.KeyFile, info.ArchLinux.Signature.KeyPassphrase, info.ArchLinux.Signature.KeyID)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
+		}
+	}
+	return sig, nil
+}
+
+func createSignatureFile(info *nfpm.Info, sig []byte) error {
+	sigFile, err := os.Create(info.Target + ".sig")
+	if err != nil {
+		return fmt.Errorf("create signature file: %w", err)
+	}
+	defer sigFile.Close()
+
+	_, err = sigFile.Write(sig)
+	if err != nil {
+		return fmt.Errorf("write signature to file: %w", err)
+	}
+
+	return nil
 }
 
 func writeKVPairs(w io.Writer, pairs map[string]string) error {
--- a/arch/arch_test.go
+++ b/arch/arch_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/goreleaser/nfpm/v2"
 	"github.com/goreleaser/nfpm/v2/files"
+	"github.com/goreleaser/nfpm/v2/internal/sign"
 	"github.com/klauspost/compress/zstd"
 	"github.com/klauspost/pgzip"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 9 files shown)


arch/arch.go
```diff
@@ -134,10 +136,8 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	defer zw.Close()
 
 	tw := tar.NewWriter(zw)
-	defer tw.Close()
```
caarlos0, line 141: this will not properly cleanup if an error happens 🤔

internal/sign/pgp.go
```diff
@@ -48,6 +48,31 @@ func PGPSignerWithKeyID(keyFile, passphrase string, hexKeyID *string) func([]byt
 	}
 }
 
+// PGPDetachedSignWithKeyID creates a detached signature.
+func PGPDetachedSignWithKeyID(message io.Reader, keyFile, passphrase string, hexKeyID *string) ([]byte, error) {
+	keyID, err := parseKeyID(hexKeyID)
```
flecno, line 33: This duplicates the code from PGPSignerWithKeyID.

- [ ] refactor PGPSignerWithKeyID and call this function

flecno:
@Elara6331 Do you want to take a look at this?

flecno:
@caarlos0 Is there any chance to get this reviewed and merged? It's a very straight forward change

internal/sign/pgp.go
```diff
@@ -20,32 +20,37 @@ import (
...
 	return func(data []byte) ([]byte, error) {
-		keyID, err := parseKeyID(hexKeyID)
-		if err != nil {
-			return nil, fmt.Errorf("%v is not a valid key id: %w", hexKeyID, err)
-		}
-
-		key, err := readSigningKey(keyFile, passphrase)
+		sig, err := PGPArmoredDetachSignWithKeyID(bytes.NewReader(data), keyFile, passphrase, hexKeyID)
 		if err != nil {
 			return nil, &nfpm.ErrSigningFailure{Err: err}
```
caarlos0, line 30: changing the error types returned is a breaking change. I don't think its really necessary...

internal/sign/pgp.go
```diff
@@ -20,32 +20,37 @@ import (
...
-		if err := openpgp.DetachSign(
-			&signature,
-			key,
-			bytes.NewReader(data),
-			&packet.Config{
-				SigningKeyId: keyID,
-				DefaultHash:  crypto.SHA256,
-			},
-		); err != nil {
-			return nil, &nfpm.ErrSigningFailure{Err: err}
```
caarlos0, line 44: same here

arch/arch.go
```diff
@@ -134,10 +136,8 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	defer zw.Close()
```
caarlos0, line 138: this will not properly cleanup if an error happens 🤔

arch/arch.go
```diff
@@ -157,7 +157,32 @@ func (ArchLinux) Package(info *nfpm.Info, w io.Writer) error {
 		return fmt.Errorf("create mtree: %w", err)
 	}
 
-	return createScripts(info, tw)
+	err = createScripts(info, tw)
+	if err != nil {
```
caarlos0:
```suggestion
	if err := createScripts(info, tw); err != nil {
```

caarlos0 (commented):
looking good overall, a couple of comments though.

thanks for the PR 🙏🏻

flecno:
Thanks! I'll go over it in the next few days

internal/sign/pgp.go
flecno: I didn't change the error type here. This is the helper function used as RPM signer that wraps the error. Maybe we should extract it to the RPM module?

flecno:
@caarlos0 I rebased the branch, fixed your comments and the lint errors and added an acceptance test for signed Arch Linux packages

arch/arch.go
```diff
@@ -403,6 +432,45 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 	}, nil
 }
 
+func createSignature(info *nfpm.Info) ([]byte, error) {
+	f, err := os.Open(info.Target)
```
caarlos0, line 436: `Package` signs the bytes it reads back from `info.Target`, but the contract of `Packager.Package(info, w)` is that the package goes to `w`. Nothing ties the two together.

Two reachable outcomes:

**1. `Target` unset — loud failure.** Any library caller gets `signing error: create signature: open package for signing: open : no such file or directory`. goreleaser's nfpm pipe is exactly this shape: it opens its own file and calls `packager.Package(info, w)` without ever assigning `info.Target`. Nothing is broken today only because goreleaser's `NFPMArchLinux` has no `signature` field yet — but that means this feature is unusable outside the nfpm CLI, and it breaks the moment goreleaser plumbs it through.

**2. `Target` set but not the same bytes as `w` — silent, and worse.** A buffered writer, or a caller that writes to a temp path and renames afterwards, makes `os.Open(info.Target)` read stale or partial data. The result is a valid PGP signature over bytes that are not the package, with no error and no warning. deb, rpm and apk cannot do this because they sign the bytes they produce.

`Target` is undocumented (`nfpm.go:328`, `yaml:"-" json:"-"`, no comment) and assigned only at `internal/cmd/package.go:109`. This PR turns it into a silent precondition of a signing feature.

**Smallest fix:** sign what you write. When a signature is configured, wrap the writer once with `io.MultiWriter(w, &buf)` and sign `buf`. deb and rpm already buffer the whole package in memory, so this is consistent and costs nothing new. Keep `Target` only to name the `.sig` file, and return an explicit error when it is empty rather than letting `os.Open("")` speak for you.

**Test:** `Default.Package(info, &bytes.Buffer{})` with a *valid* key file and `Target` unset, asserting a specific, actionable error.

<sub>Correction: an earlier revision of this comment quoted the error with `create signature:` doubled. On the `os.Open` path it appears once, as shown above; the doubling happens on the signing-failure path (see the comment on `arch/arch.go:442-453`).</sub>

arch/arch_test.go
```diff
@@ -209,6 +210,59 @@ func TestArchOverrideArchitecture(t *testing.T) {
...
+func TestArchSignatureError(t *testing.T) {
+	info := exampleInfo()
+	info.ArchLinux.Signature.KeyFile = "/does/not/exist"
+
+	var pkg bytes.Buffer
+	err := Default.Package(info, &pkg)
+	require.Error(t, err)
+
+	var expectedError *nfpm.ErrSigningFailure
+	require.ErrorAs(t, err, &expectedError)
```
caarlos0, line 242: This test passes for the wrong reason.

`exampleInfo()` never sets `Target`, and `createSignature` opens `info.Target` (`arch.go:436`) *before* it ever reads `KeyFile` (`arch.go:451`). So the error under test is `open : no such file or directory`, not anything to do with `/does/not/exist`.

Two consequences: the test would pass unchanged if you swapped in a **valid** key file, and it would pass if `sign.PGPDetachedSignWithKeyID` were deleted outright. It asserts nothing about key handling.

**Fix:** set `info.Target` to a real temp file the way the other two tests do, then tighten the assertion so it can only pass for the intended reason:

```go
require.ErrorContains(t, err, "/does/not/exist")
```

Then add the empty-`Target` case as its own test, per the comment on `arch/arch.go:436`.

arch/arch_test.go
```diff
@@ -209,6 +210,59 @@ func TestArchOverrideArchitecture(t *testing.T) {
...
+	t.Cleanup(func() { require.NoError(t, f.Close()) })
+	info.Target = f.Name()
+	require.NoError(t, Default.Package(info, f))
+
+	signature, err := os.ReadFile(info.Target + ".sig")
+	require.NoError(t, err)
+
+	_, err = f.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+	err = sign.PGPVerify(f, signature, "../internal/sign/testdata/pubkey.asc")
```
caarlos0, line 229: Nothing in the repository asserts that the signature is **binary** — which is the entire reason `PGPDetachedSignWithKeyID` was added instead of reusing the existing `PGPArmoredDetachSignWithKeyID`.

`PGPVerify` branches on `isASCII(signature)` (`internal/sign/pgp.go:143`) and happily accepts either form. `TestArchSignatureCallback` just below deliberately returns an **armored** signature and passes. `pacman-key --verify` shells out to gpg, which also auto-detects armor. So swapping the binary signer for the armored one would pass every test in this repo while emitting a `.sig` that is not the format pacman expects.

One line closes it:

```go
require.False(t, bytes.HasPrefix(signature, []byte("-----BEGIN")))
```

Minor, same test: line 218 passes `info.Target` to `os.CreateTemp` as the name pattern, but `info.Target` is `""` at that point. A literal like `"pkg"` says what is actually meant.

testdata/acceptance/archlinux.dockerfile
```diff
@@ -75,8 +75,12 @@ RUN test ! -d /usr/foo/bar/something
...
 # ---- signed test ----
 FROM min AS signed
-RUN echo "Arch Linux has no signature support"
-
+COPY ${package}.sig /tmp/foo.pkg.tar.zst.sig
+COPY keys/pubkey.asc /tmp/pubkey.asc
+RUN pacman-key --init
+RUN pacman-key --add /tmp/pubkey.asc
+RUN pacman-key --lsign-key 866F6C83BAB3E49381ADE4C1BC8ACDD415BD80B3
+RUN pacman-key --verify /tmp/foo.pkg.tar.zst.sig 2>&1 | grep "gpg: Good signature from \"nfpm test key <test@example.com>\" \[full\]"
```
caarlos0, line 83: This stage proves that **gpg** accepts the signature, but never that **pacman** does. `signed` is `FROM min`, so `pacman -U` already ran a layer earlier — before the `.sig` was copied in — and therefore never saw it.

Per `pacman.conf(5)`, the Arch default `LocalFileSigLevel = Optional` means "signatures are checked if present; absence of a signature is not an error. An invalid signature is a fatal error." So one extra line after the `lsign-key` gives real end-to-end coverage without touching any config, and it is the only check here that would catch an armored-vs-binary regression at the libalpm level:

```dockerfile
RUN pacman --noconfirm -U /tmp/foo.pkg.tar.zst
```

arch/arch.go
```diff
@@ -403,6 +432,45 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
...
+		sig, err = sign.PGPDetachedSignWithKeyID(data, info.ArchLinux.Signature.KeyFile, info.ArchLinux.Signature.KeyPassphrase, info.ArchLinux.Signature.KeyID)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
+		}
+	}
+	return sig, nil
+}
+
+func createSignatureFile(info *nfpm.Info, sig []byte) error {
+	sigFile, err := os.Create(info.Target + ".sig")
```
caarlos0, line 460: The `.sig` is a second output file that nothing ever cleans up.

- `internal/cmd/package.go:112` does `os.Remove(target)` when `Package` fails, but never `os.Remove(target + ".sig")`.
- Deterministic footgun: build a signed package, then rebuild the same target with `key_file` removed. The package is replaced; the old `.sig` survives beside it and now covers different bytes. Since `LocalFileSigLevel = Optional` treats an invalid signature as fatal, `pacman -U` rejects the new package with nothing pointing at the cause.

**Fix:** remove `target + ".sig"` in the CLI error path, and remove a pre-existing `.sig` here when no signature is configured.

**Test:** in `internal/cmd`, create a `.sig`, run a package command with signing disabled, assert the `.sig` is gone.

arch/arch.go
```diff
@@ -403,6 +432,45 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
...
+	if err != nil {
+		return fmt.Errorf("create signature file: %w", err)
+	}
+	defer sigFile.Close()
+
+	_, err = sigFile.Write(sig)
+	if err != nil {
+		return fmt.Errorf("write signature to file: %w", err)
+	}
+
```
caarlos0, line 470: This helper is `os.WriteFile`:

```go
return os.WriteFile(info.Target+".sig", sig, 0o644)
```

One line instead of twelve, an explicit mode rather than whatever umask gives you, and — unlike `defer sigFile.Close()` — it reports a `Close` failure instead of discarding it.

arch/arch.go
```diff
@@ -403,6 +432,45 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
...
+	var sig []byte
+	if signFn := info.ArchLinux.Signature.SignFn; signFn != nil {
+		sig, err = signFn(data)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
+		}
+	} else {
+		sig, err = sign.PGPDetachedSignWithKeyID(data, info.ArchLinux.Signature.KeyFile, info.ArchLinux.Signature.KeyPassphrase, info.ArchLinux.Signature.KeyID)
+		if err != nil {
+			return nil, fmt.Errorf("create signature: %w", err)
```
caarlos0, line 453: Two nits in this function.

`bufio.NewReader(f)` costs more than it saves. `openpgp.DetachSign` copies through `io.Copy`, which already uses a 32 KiB buffer, so a 4 KiB `bufio.Reader` only adds a copy and forces smaller reads. Pass `f` directly.

The `"create signature: "` prefix is applied here **and** again by the caller at `arch.go:179`, so a real failure reads `signing error: create signature: create signature: detach sign: reading PGP key file: ...`. Drop the inner wrap and return the error unchanged — the caller already names the operation.

nfpm.go
```diff
@@ -268,6 +271,11 @@ func (c *Config) expandEnvVars() {
 		c.APK.Signature.KeyPassphrase = apkPassphrase
 	}
 
+	archlinuxPassphrase := os.Expand("$NFPM_ARCHLINUX_PASSPHRASE", c.envMappingFunc)
+	if archlinuxPassphrase != "" {
+		c.ArchLinux.Signature.KeyPassphrase = archlinuxPassphrase
+	}
```
caarlos0, line 277: `nfpm_test.go` covers the global passphrase plus the per-format override for deb, rpm and apk, but this new branch has no unit test. One assertion added to the existing table keeps `NFPM_ARCHLINUX_PASSPHRASE` and its `$NFPM_PASSPHRASE` fallback from regressing silently.

testdata/acceptance/core.signed.yaml
```diff
@@ -19,3 +19,6 @@ rpm:
 apk:
   signature:
     key_file: ./internal/sign/testdata/rsa_unprotected.priv
+archlinux:
+  signature:
+    key_file: ./internal/sign/testdata/privkey_unprotected.asc
```
caarlos0, line 24: Missing trailing newline (`\ No newline at end of file` in the diff). Worth fixing, especially since this same PR *adds* one to `www/static/schema.json`.

www/content/docs/configuration.md
```diff
@@ -513,6 +513,18 @@ archlinux:
...
+  signature:
+    # PGP secret key, the passphrase is taken from the environment variable $NFPM_ARCHLINUX_PASSPHRASE with
+    # a fallback to $NFPM_PASSPHRASE.
+    # This will expand any env var you set in the field, e.g. key_file: ${SIGNING_KEY_FILE}
+    key_file: key.gpg
+
+    # PGP secret key id in hex format, if it is not set it will select the first subkey
+    # that has the signing flag set. You may need to set this if you want to use the primary key as the signing key.
+    # This will expand any env var you set in the field, e.g. key_id: ${ARCHLINUX_SIGNING_KEY_ID}
+    key_id: bc8acdd415bd80b3
```
caarlos0, line 526: This block never tells the reader that archlinux signing produces a **second file**, `<package>.sig`, or that pacman needs it sitting next to the package. That is the user-visible novelty here — every other format embeds its signature — so it deserves an explicit line.

Worth adding too: a `SignFn` must return a **binary** detached signature, not an ASCII-armored one. The `SignFn` doc comment in `nfpm.go` now says archlinux receives "the full package content", but says nothing about the expected output format.

caarlos0 (changes requested):
Thanks for this. The `nfpm package` CLI path works, and I verified a few things independently so they do not come up again:

- `866F6C83BAB3E49381ADE4C1BC8ACDD415BD80B3` really is the primary key of both `internal/sign/testdata/privkey_unprotected.asc` and `testdata/acceptance/keys/pubkey.asc`, and the UID is exactly `nfpm test key <test@example.com>`, so the acceptance grep matches.
- The `.sig` does reach the Docker build context: `acceptance_test.go` sets `info.Target` before `pkg.Package`, and the build context is `testdata/acceptance`.
- The double `Close()` of the tar and zstd writers is safe. `archive/tar` returns `nil` on the second call via its `ErrWriteAfterClose` guard, and `klauspost/compress@v1.18.6` returns `nil` via `errors.Is(s.err, ErrEncoderClosed)`. Neither emits extra bytes, including on the `fullFrameWritten` path.
- `TestArchSignature` genuinely guards the new close ordering: without the explicit `tw.Close()`/`zw.Close()` the signed bytes are a truncated prefix and `PGPVerify` fails.

Requesting changes for two things:

1. The packager signs bytes it reads back from `info.Target` instead of the bytes it wrote to `w`. In one variant this fails loudly for any library caller; in another it silently signs the wrong bytes. See the comment on `arch/arch.go:436`.
2. `TestArchSignatureError` passes for the wrong reason and asserts nothing about key handling. See the comment on `arch/arch_test.go:233-242`.

The rest are smaller: one real coverage gap in the acceptance stage, a missing binary-vs-armored assertion, stale `.sig` cleanup, and some nits.

**Verification gap, stated plainly:** I was asked not to build or run tests locally. Every claim here comes from reading the source, reading the module sources in `GOMODCACHE`, running `gpg` against the test keys, and `pacman.conf(5)`. I have *not* empirically demonstrated that `TestArchSignatureError` passes for the wrong reason, and I have not shown that any suggested regression test fails without its fix. Please confirm both before acting on them.

*Reviewed by an AI agent (GitHub Copilot CLI) at a maintainer's request.*

PR #1072, opened 2026-04-15
fix: include RPM doc/licence/readme files in non-RPM packages

### LINKED ISSUE #1037: type: doc excludes files from non-RPM packages instead of treating them as   normal files
### What happened?

Files with type: doc (and other RPM-specific types like license, readme) are completely
  excluded from deb and apk packages, contradicting the documentation which states they
  should be "ignored" (treated as normal files) by other packagers.

  Documentation vs Actual Behavior

  Documentation states (https://pkg.go.dev/github.com/goreleaser/nfpm/v2/files,
  https://nfpm.goreleaser.com/docs/configuration/):

  TypeRPMDoc is "the type of an RPM doc file which is ignored by other packagers."

  "For packagers other than RPM, [...] the remaining RPM-specific types are the same as '',
   a normal file."

  Actual behavior in files/files.go (isRelevantForPackager function):

  if packager != "rpm" &&
      (content.Type == TypeRPMDoc || content.Type == TypeRPMLicence ||
          content.Type == TypeRPMLicense || content.Type == TypeRPMReadme ||
          content.Type == TypeRPMGhost) {
      return false  // Files are EXCLUDED, not treated as normal files
  }

 Users who want to mark documentation files with %doc in RPM while still including them in
   deb/apk packages cannot do so. The workaround is to not use type: doc at all, which
  means losing the %doc functionality in RPM.

nfpm.yaml:
  contents:
    - src: ./README.md
      dst: /usr/share/doc/mypackage/README.md
      type: doc
      file_info:
        mode: 0644

  Expected behavior:
  - RPM: File included with %doc marking
  - Deb/APK: File included as a normal file (type ignored)

  Actual behavior:
  - RPM: File included with %doc marking ✓
  - Deb/APK: File is completely missing from package ✗

### nfpm version

```bash
latest
```

### Search

- [x] I did search for other open and closed issues before opening this.

### Code of Conduct

### Additional context

  In isRelevantForPackager, instead of returning false for RPM-specific types on non-RPM
  packagers, the types should be normalized to regular files:

  // Option 1: Don't filter, let them fall through as regular files
  // Remove the filtering block entirely, or:

  // Option 2: Normalize the type instead of excluding
  if packager != "rpm" {
      switch content.Type {
      case TypeRPMDoc, TypeRPMLicence, TypeRPMLicense, TypeRPMReadme:
          content.Type = TypeFile  // Treat as normal file
      case TypeRPMGhost:
          return false  // Ghost files should still be excluded
      }
  }

  - Source: https://github.com/goreleaser/nfpm/blob/main/files/files.go
  - Docs: https://pkg.go.dev/github.com/goreleaser/nfpm/v2/files

Yanhu007:
Fixes #1037

## Problem

Files with `type: doc`, `type: licence`, `type: license`, or `type: readme` are completely excluded from deb and apk packages. The documentation states these RPM-specific types should be "ignored" (treated as normal files) by other packagers.

`isRelevantForPackager` returns `false` for all these types when the packager is not "rpm", causing them to be filtered out entirely.

## Fix

Only exclude `TypeRPMGhost` from non-RPM packagers (ghost files have no actual content, so exclusion makes sense). Doc/licence/license/readme types are no longer excluded, allowing them to be treated as normal files per the documentation.

This matches the documented behavior:
> "For packagers other than RPM, [...] the remaining RPM-specific types are the same as '', a normal file."

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -335,11 +335,14 @@ func isRelevantForPackager(packager string, content *Content) bool {
 	}
 
 	if packager != "rpm" &&
-		(content.Type == TypeRPMDoc || content.Type == TypeRPMLicence ||
-			content.Type == TypeRPMLicense || content.Type == TypeRPMReadme ||
-			content.Type == TypeRPMGhost) {
+		content.Type == TypeRPMGhost {
+		// Ghost files only apply to RPM; exclude from other packagers.
 		return false
 	}
+
+	// RPM-specific document types (doc, licence, license, readme) are treated
+	// as normal files by non-RPM packagers per documentation, so they are NOT
+	// excluded here.
 
 	if packager != "deb" && content.Type == TypeDebChangelog {
 		return false


caarlos0:
this have tests guarding it - i think it may be that we need to update the docs instead 🤔

Issue #1079, opened 2026-04-30
Segmentation fault if `overrides.<pkg>` is empty in nfpm.yaml


dleske:
### What happened?

If `overrides.$pkg` is defined but empty in the YAML configuration, this causes a segmentation fault.  `overrides` itself can be empty.

I would expect the nonsensical clause to be ignored or to receive a warning.

### How can we reproduce this?

Have the following or similar in the config file:

```
overrides:
  deb:
    # nothing defined here
# return to rest of config
```

### nfpm version

```bash
Built from source today. Given version is `dev`.

Full output anyway:


       _____ ____  __  __
 _ __ |  ___|  _ \|  \/  |
| '_ \| |_  | |_) | |\/| |
| | | |  _| |  __/| |  | |
|_| |_|_|   |_|   |_|  |_|
nfpm: a simple and 0-dependencies apk, arch linux, deb, ipk, msix, and rpm packager written in Go
https://nfpm.goreleaser.com

GitVersion:    dev
GitCommit:     unknown
GitTreeState:  unknown
BuildDate:     unknown
BuiltBy:       unknown
GoVersion:     go1.25.9
Compiler:      gc
ModuleSum:     h1:sMNcGjbVnABe33avsFo++WwKubY8msP5j9ieulX09VI=
Platform:      linux/amd64
```

### Search

- [x] I did search for other open and closed issues before opening this.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Additional context

_Thank you for the software_

dleske:
I have a PR I will submit shortly.

PR #1080, opened 2026-04-30
fix: tolerate empty overrides packager clause

dleske:
Addresses #1079.

If an `overrides.$packager` clause is empty in the configuration, this is detected and a warning issued instead of producing a segmentation violation.

I only dabble in Go from time to time so apologies if the code isn't up to standard.

### DIFF
--- a/nfpm.go
+++ b/nfpm.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"log"
 	"os"
 	"slices"
 	"sort"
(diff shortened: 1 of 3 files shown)


nfpm.go
```diff
@@ -211,13 +212,17 @@ func (c *Config) expandEnvVars() {
...
+			log.Printf("warning: override '%s' is empty; ignoring", or)
+		} else {
+			c.Overrides[or].Conflicts = c.expandEnvVarsStringSlice(c.Overrides[or].Conflicts)
+			c.Overrides[or].Depends = c.expandEnvVarsStringSlice(c.Overrides[or].Depends)
+			c.Overrides[or].Replaces = c.expandEnvVarsStringSlice(c.Overrides[or].Replaces)
+			c.Overrides[or].Recommends = c.expandEnvVarsStringSlice(c.Overrides[or].Recommends)
+			c.Overrides[or].Provides = c.expandEnvVarsStringSlice(c.Overrides[or].Provides)
+			c.Overrides[or].Suggests = c.expandEnvVarsStringSlice(c.Overrides[or].Suggests)
+			c.Overrides[or].Contents = c.expandEnvVarsContents(c.Overrides[or].Contents)
+		}
```
caarlos0, line 216: i don't think we need to log, IMO we can just continue:

```suggestion
		if c.Overrides[or] == nil {
			continue
		}
		c.Overrides[or].Conflicts = c.expandEnvVarsStringSlice(c.Overrides[or].Conflicts)
		c.Overrides[or].Depends = c.expandEnvVarsStringSlice(c.Overrides[or].Depends)
		c.Overrides[or].Replaces = c.expandEnvVarsStringSlice(c.Overrides[or].Replaces)
		c.Overrides[or].Recommends = c.expandEnvVarsStringSlice(c.Overrides[or].Recommends)
		c.Overrides[or].Provides = c.expandEnvVarsStringSlice(c.Overrides[or].Provides)
		c.Overrides[or].Suggests = c.expandEnvVarsStringSlice(c.Overrides[or].Suggests)
		c.Overrides[or].Contents = c.expandEnvVarsContents(c.Overrides[or].Contents)
```

caarlos0:
Thanks for the PR - one small comment but otherwise LGTM

nfpm.go
dleske: I like the warning because it signals a possible mistake by the user so they can correct it, but that's just my opinion and it's your project. I made the suggested change. :)

dleske:
Ready for another review.

caarlos0 (approved):

caarlos0:
Thank you!

PR #1081, opened 2026-05-01
srpm: set rpmpack SourceRPM metadata flag

alexma233:
When building source RPMs, the `SOURCERPM` (RPMTAG_SOURCERPM, 1044) tag must not be present.  RPM utilities treat its presence as indicating a binary (non-source) RPM.  This causes build services like Fedora COPR to reject source RPMs generated by the `srpm` packager.

The upstream rpmpack (google/rpmpack#105) adds a `SourceRPM` field to `RPMMetaData` that controls whether `SOURCERPM` is written.  This PR:

- Sets `meta.SourceRPM = true` when the packager format is `srpm`
- Uses a `replace` directive pointing to the rpmpack fork with the new field

### Dependency

Depends on https://github.com/google/rpmpack/pull/105 being merged and released.

The `replace` directive in `go.mod` should be removed and the rpmpack dependency bumped once a new release is available.

### Verification

- `go test ./rpm/...` — 73 tests pass, including `TestSRPM`
- Generated SRPM no longer contains the `SOURCERPM` tag header

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -95,3 +95,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	software.sslmate.com/src/go-pkcs12 v0.7.1 // indirect
 )
+
+replace github.com/google/rpmpack => github.com/alexma233/rpmpack v0.7.2-0.20260501025250-ba11ef02e94a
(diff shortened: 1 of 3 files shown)


caarlos0:
Good catch! Thank you!

Let's see if rpmpack merges it soon 🙏🏻

alexma233:
https://github.com/goreleaser/nfpm/pull/1104

PR #1088, opened 2026-05-16
fix: improve errors

otjdiepluong:
Summary:
- Tighten error wrapping in the deb/rpm writer paths so callers can inspect the original cause.
- Add focused coverage for the wrapped-error behavior.

Notes:
- I kept this scoped to the relevant packaging path and tests.

### DIFF
--- a/deb/deb_test.go
+++ b/deb/deb_test.go
@@ -1660,3 +1660,55 @@ func verifyDpkgSigFileHashes(arFiles map[string][]byte, msg string) error {
 	}
 	return nil
 }
+
+type failAfterNWriter struct {
+	n   int
+	err error
+}
+
+func (w *failAfterNWriter) Write(p []byte) (int, error) {
+	if w.n <= 0 {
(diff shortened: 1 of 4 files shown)


deb/deb_test.go
```diff
@@ -1660,3 +1660,55 @@ func verifyDpkgSigFileHashes(arFiles map[string][]byte, msg string) error {
...
+	if w.n <= 0 {
+		return 0, w.err
+	}
+	if len(p) > w.n {
+		p = p[:w.n]
+	}
+	w.n -= len(p)
+	return len(p), nil
+}
+
```
caarlos0, line 1679: we don't really need fail after n bytes on the tests I think?

could be just

```go
type errWriter struct{ err error }

func (e *errWriter) Write([]byte) (int, error) { return 0, e.err }
```

caarlos0 (commented):
one comment, looks good otherwise

thanks 🙏🏻

PR #1061, opened 2026-03-28
feat: xbps packager support

actx4gh:
## xbps packager support

Resolves #718

### What this adds

Native XBPS package generation for [Void Linux](https://voidlinux.org/).
Packages are zstd-compressed tarballs containing `props.plist`, `files.plist`,
lifecycle scripts (`INSTALL`/`REMOVE`), and content entries, matching the
format consumed by `xbps-rindex`, `xbps-install`, and related tools.

### Supported features

- Architecture mapping (amd64, arm64, arm6, arm7, 386, noarch, plus explicit `xbps.arch` override)
- Dependencies (`run_depends`), conflicts, provides, replaces, reverts
- Config files (all config variants recorded in `conf_files` metadata)
- Lifecycle scripts (pre/post install/remove wrapped into XBPS `INSTALL`/`REMOVE` dispatch)
- Alternatives (`xbps.alternatives`)
- Preserve mode (`xbps.preserve`)
- Tags (`xbps.tags`)
- Short and long descriptions
- Environment variable expansion for all XBPS-specific fields

### How scripts work

XBPS uses a single `INSTALL` script called with positional args (`$1` = pre/post,
`$4` = yes/no for update). nfpm's four separate script hooks are wrapped into
this dispatch via heredoc extraction to a temp directory and a `case` block.
Same approach for `REMOVE`.

### Testing

- Unit tests in `xbps/xbps_test.go` covering naming, metadata, plist serialization,
  script rendering, and error paths
- Docker-based acceptance tests on `ghcr.io/void-linux/void-glibc-full`:
  - **lifecycle**: install, query, reconfigure, remove (with config preservation)
  - **upgrade**: v1 to v2 with script semantic args and metadata assertions
  - **preserve**: package-level `preserve` flag across upgrades
  - **metadata**: noarch mapping and alternatives wiring
- Taskfile smoke test for local `build + install + query` loop

### Scope and known limitations

- CI matrix tests **x86_64 only**. Arch mapping covers other targets but they
  are not exercised in acceptance tests yet.
- No package signing support (XBPS repos can use unsigned local repos, which
  is what the acceptance tests do).

### Changes to shared code

- `nfpm.go`: Added `XBPS` struct, `hasAnyArch()` replacing inline arch check
  (now correctly considers all packager-specific arch overrides, not just deb/rpm/apk)
- `acceptance_test.go`: Extracted `acceptWithOldPackage()` and `acceptPackageTarget()`
[...]

PR #1094, opened 2026-05-27
feat: Add XBPS packager support

### LINKED ISSUE #718: Support xbps packages

nFPM doesn't have support for the XBPS package format used by Void Linux

Implement support for XBPS packages. I can work on a PR for this.

N/A

### Search

### Code of Conduct

### Additional context

_No response_

actx4gh:
## Problem

nFPM does not currently support generating XBPS packages.

Resolves #718.

This PR replaces the earlier closed PR #1061 with a rebuilt branch, a narrower commit history, package signature sidecar support, and additional validation against Void Linux tooling.

## Change summary

This adds native XBPS package generation to nFPM.

The new `xbps` packager writes `.xbps` package archives directly in Go rather than shelling out to `xbps-create`. The generated package is a zstd-compressed tar archive containing the XBPS metadata files, optional lifecycle scripts, and the package payload entries consumed by normal XBPS tooling.

Supported package behavior includes:

- architecture mapping, with explicit `xbps.arch` override support;
- conventional XBPS filenames and `.xbps` extension support;
- generic nFPM metadata mapped into XBPS package metadata;
- dependencies, conflicts, provides, replaces, and XBPS reverts;
- config-file metadata derived from existing nFPM content types;
- `xbps.short_desc`, `xbps.preserve`, `xbps.tags`, and `xbps.alternatives`;
- install and remove lifecycle script wrapping through XBPS `INSTALL` and `REMOVE` action dispatch;
- optional adjacent package signature sidecar output through `xbps.signature.key_file`.

As such, the implementation follows nFPM’s existing first-class packager model while keeping XBPS-specific archive and metadata behavior inside the new `xbps` package.

## Signing

This PR includes XBPS package signature sidecar support.

When `xbps.signature.key_file` is configured, nFPM signs the generated package digest with RSA SHA-256 and writes an adjacent `<package>.xbps.sig2` file.

This is intentionally limited to package sidecar signing. It does not create or sign repository metadata, publish repositories, manage remote repositories, or orchestrate `xbps-rindex --sign`. Repository indexing and repository signing remain separate XBPS tooling workflows.

## Runtime validation

XBPS behavior is validated primarily against Void Linux, because Void is the main public environment where this package format is exercised.

The acceptance tests use a local XBPS repository flow:

1. generate `.xbps` packages with nFPM;
2. index the local repository with `xbps-rindex -a`;
3. install from that repository with XBPS tooling;
4. query package metadata;
5. reconfigure the package;
6. upgrade the package;
7. remove the package.

Covered XBPS scenarios include install, metadata/query behavior, lifecycle scripts, reconfigure, remove, upgrade, noarch packages, preserve behavior, alternatives, tags, reverts, and config metadata.

I also performed additional runtime testing in my Void/s6 VM environment using packages from my `void-s6` repository:

https://codeberg.org/aaron_colichia/void-s6

That testing gave additional confidence that the generated packages work outside the unit-test-only path and can be consumed in a real Void-based environment.

## Validation

Local validation run for the rebuilt branch:

- `go test ./internal/sign ./xbps .`
- `go test ./...`
- `go test -tags=acceptance -run '^TestXBPSSpecific$' .`

A fork-local GitHub Actions preflight passed for the branch-specific checks.

Two fork-local failures were baseline or toolchain issues outside this XBPS work:

- `license-check`: `go-licenses/v2@latest` rejects the existing dependency `github.com/cyphar/filepath-securejoin` as `MPL-2.0`.
- `docs / build`: Hugo `latest` pulled a newer Hugo version that fails in the Hextra RSS template on `.Site.Author.email`.

The only branch-specific lint issue found during preflight was fixed with `gofumpt`.

## Non-goals

This PR does not add:

- XBPS repository publishing;
- XBPS repository metadata signing;
- remote repository management;
- `xbps-rindex --sign` orchestration;
- `xbps-create` as the production packaging backend;
- `xbps-src` or void-packages template generation;
- automatic shared-library provide/requires discovery;
- purge lifecycle hook support;
- a broad cross-packager signing redesign.

## Note

I used AI assistance to help scaffold and organize the shape of this PR, then reviewed the code and performed the validation above. I can answer implementation questions about the changes.


testdata/acceptance/xbps.dockerfile
```diff
@@ -0,0 +1,73 @@
+FROM ghcr.io/void-linux/void-glibc-full:latest AS scenario
```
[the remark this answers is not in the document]
actx4gh: archlinux, openwrt/rootfs (ipk), and fedora all use :latest.  leaving it like this. reopen if you disagree.

internal/sign/rsa.go
```diff
@@ -47,45 +72,29 @@ func RSASignSHA1Digest(sha1Digest []byte, keyFile, passphrase string) ([]byte, e
...
 		if err != nil {
 			return nil, fmt.Errorf("parse PKCS#8 private key: %w", err)
 		}
-		privTmp, ok := privAny.(crypto.Signer)
+		priv, ok := privAny.(crypto.Signer)
 		if !ok {
 			return nil, fmt.Errorf("cannot sign with given private key")
 		}
-		priv = privTmp
+		return priv, nil
```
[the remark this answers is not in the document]
actx4gh: assert *rsa.PrivateKey and return the existing errNoRSAKey

www/content/docs/quick-start.md
```diff
@@ -47,9 +47,33 @@ Use [`nfpm package`](/docs/cmd/nfpm_package) to create your packages:
 nfpm pkg --packager deb --target /tmp/
 nfpm pkg --packager rpm --target /tmp/
 nfpm pkg --packager apk --target /tmp/
+nfpm pkg --packager xbps --target /tmp/
 ```
 
-You can also use `ipk`, `archlinux`, and `msix` as packagers.
+You can also use `ipk`, `archlinux`, `msix`, and `xbps` as packagers.
```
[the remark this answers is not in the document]
actx4gh: reverted second mention

nfpm.go
```diff
@@ -281,6 +281,27 @@ func (c *Config) expandEnvVars() {
...
+	if xbpsPassphrase != "" {
+		c.XBPS.Signature.KeyPassphrase = xbpsPassphrase
+	}
+	nfpmXBPSPassphrase := os.Expand("$NFPM_XBPS_PASSPHRASE", c.envMappingFunc)
+	if nfpmXBPSPassphrase != "" {
+		c.XBPS.Signature.KeyPassphrase = nfpmXBPSPassphrase
+	}
+
 	// RPM specific
 	c.RPM.Packager = os.Expand(c.RPM.Packager, c.envMappingFunc)
```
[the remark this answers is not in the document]
actx4gh: @caarlos0  appears to have existed prior to the PR, but smell looks valid so addressed. please confirm or let me know if you want to back this out.

xbps/xbps.go
```diff
@@ -0,0 +1,740 @@
...
+}
+
+func fileEntry(content *files.Content) (plistDict, error) {
+	entry := plistDict{
+		"file": files.NormalizeAbsoluteFilePath(content.Destination),
+	}
+	if content.Type == files.TypeSymlink {
+		entry["target"] = normalizeTargetForMetadata(content.Destination, content.Source)
+		return entry, nil
+	}
```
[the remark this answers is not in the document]
actx4gh: Make the symlink target in package metadata match the actual symlink written in the payload.

actx4gh:
Still here if you need me!

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+
+	parts := []string{base}
+	if pre := normalizeVersionPart(info.Prerelease); pre != "" {
+		parts = append(parts, pre)
+	}
+	if meta := normalizeVersionPart(info.VersionMetadata); meta != "" {
+		parts = append(parts, meta)
+	}
+	return strings.Join(parts, ".")
+}
```
caarlos0, line 84: **Blocker: a hyphen in the version silently produces a package XBPS resolves under the wrong name.**

`normalizeVersionPart` only *trims* leading/trailing hyphens, so inner hyphens survive into `version()` and then into `pkgver()`.

Upstream `xbps_pkg_name()` (`lib/util.c`) splits the pkgver at the **last** hyphen via `strrchr(pkg, '-')`, and `xbps-rindex` keys the repository index off that derived name, not off the `pkgname` you write into `props.plist` (`bin/xbps-rindex/index-add.c:226` calls `xbps_pkg_name(pkgname, sizeof(pkgname), pkgver)`).

Trigger - this is the packager's own example fixture, so it is the default path, not an edge case:

```yaml
version: v1.2.3-beta-1+git-1     # semver allows '-' in prerelease and in build metadata
```

- `version()` -> `1.2.3.beta-1.git-1`
- `pkgver()` -> `foo-1.2.3.beta-1.git-1_2`
- `xbps_pkg_name()` -> **`foo-1.2.3.beta-1.git`**

`version_schema: none` with `version: 1.2.3-rc1` breaks the same way, yielding `foo-1.2.3`.

Impact: nfpm exits 0 and writes a package whose `props.plist` claims `pkgname=foo`, but `xbps-rindex -a` files it under the mangled name, so `xbps-install foo` cannot resolve it. XBPS versions may not contain `-` (name/version separator) or `_` (revision separator).

Smallest fix - replace rather than trim, and apply it to the base version too, not just the prerelease/metadata parts:

```go
func normalizeVersionPart(value string) string {
	value = strings.TrimSpace(value)
	value = strings.NewReplacer("-", ".", "_", ".").Replace(value)
	return strings.Trim(value, ".")
}
```

Test: assert `pkgver` for `Prerelease: "beta-1"` / `VersionMetadata: "git-1"` contains exactly one `-` and one `_`. That test fails today. This also requires updating the expectations in `xbps/xbps_test.go` that currently encode the broken value - see my comment there.

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+	}
+	buf.WriteString("\n</plist>\n")
+	return buf.Bytes(), nil
+}
+
+func propsManifest(info *nfpm.Info) (plistDict, error) {
+	copyInfo := *info
+	normalized, err := ensureValidArch(&copyInfo)
+	if err != nil {
+		return nil, err
```
caarlos0, line 300: **Blocker: `arch: arm6` and `arch: arm7` fail outright.**

`ensureValidArch` is not idempotent, but it is called twice on the same `Info`. `Package` normalizes in place at line 638 (`info, err = ensureValidArch(info)`), then `propsManifest` re-normalizes a copy here.

Every value in `archToXBPS` is also a key - `noarch`, `x86_64`, `i686`, `aarch64` - **except** `armv6l` and `armv7l`. So:

1. `Package` maps `arm6` -> `info.Arch = "armv6l"`
2. `propsManifest` looks up `archToXBPS["armv6l"]` -> miss
3. `nfpm pkg -p xbps` fails with `xbps: unsupported architecture "armv6l"`

`arm6`/`arm7` are documented as supported in `www/content/docs/arch-mapping.md`, so a user following the docs hits a hard failure.

Smallest fix - normalize once rather than papering over it with `armv6l`/`armv7l` self-entries. `Package` has already normalized `info` by the time `propsManifest` runs, so this function should just read `info.Arch`:

```go
func propsManifest(info *nfpm.Info) (plistDict, error) {
	pv, err := pkgver(info)
	if err != nil {
		return nil, err
	}
	manifest := plistDict{
		"architecture": info.Arch,
		...
```

`ConventionalFileName` (line 512) keeps its own copy-and-normalize, since it is called on a non-normalized `Info`.

Test: drive `Default.Package` (not `ensureValidArch`) across every key in `archToXBPS` and assert `props.plist` carries the expected architecture. The existing `TestEnsureValidArchMappings` calls the helper exactly once, so it cannot catch non-idempotency.

nfpm.go
```diff
@@ -648,6 +695,17 @@ func Validate(info *Info) (err error) {
...
+func hasAnyArch(info *Info) bool {
+	return info.Arch != "" ||
+		info.Deb.Arch != "" ||
+		info.RPM.Arch != "" ||
+		info.APK.Arch != "" ||
+		info.ArchLinux.Arch != "" ||
+		info.IPK.Arch != "" ||
+		info.XBPS.Arch != "" ||
+		info.MSIX.Arch != ""
+}
```
caarlos0, line 707: **Out of scope: this changes arch validation for deb, rpm, and apk.**

The old condition rejected only when `Arch == "" && (Deb.Arch == "" || RPM.Arch == "" || APK.Arch == "")` - i.e. valid required `Arch != ""` **or all three** of deb/rpm/apk arch set. `hasAnyArch` ORs eight fields, so **any one** now suffices.

Concretely, this was rejected before and is accepted now:

```yaml
name: foo
version: 1.2.3
msix:
  arch: x64
```

Run `nfpm pkg --packager deb` on it and `WithDefaults` silently fills `info.Arch = "amd64"` (`nfpm.go:363`), producing an amd64 deb the user never asked for. The old code failed loudly.

The loosening may well be right, but it is a deb/rpm/apk behavior change this PR does not need. Smallest fix here - add only the xbps clause and drop the helper:

```go
if info.Arch == "" && (info.Deb.Arch == "" || info.RPM.Arch == "" || info.APK.Arch == "" || info.XBPS.Arch == "") {
	return ErrFieldEmpty{"arch"}
}
```

Then open a separate PR for the general loosening, covering archlinux/ipk/msix too. The table test added at `nfpm_test.go:263` should move with it.

nfpm.go
```diff
@@ -283,6 +283,27 @@ func (c *Config) expandEnvVars() {
...
+		c.XBPS.Alternatives[i].Group = os.Expand(c.XBPS.Alternatives[i].Group, c.envMappingFunc)
+		c.XBPS.Alternatives[i].LinkName = os.Expand(c.XBPS.Alternatives[i].LinkName, c.envMappingFunc)
+		c.XBPS.Alternatives[i].Target = os.Expand(c.XBPS.Alternatives[i].Target, c.envMappingFunc)
+	}
+	c.XBPS.Signature.KeyFile = os.Expand(c.XBPS.Signature.KeyFile, c.envMappingFunc)
+	c.XBPS.Signature.KeyPassphrase = generalPassphrase
+	xbpsPassphrase := os.Expand("$XBPS_PASSPHRASE", c.envMappingFunc)
+	if xbpsPassphrase != "" {
+		c.XBPS.Signature.KeyPassphrase = xbpsPassphrase
+	}
```
caarlos0, line 301: **Please drop the unprefixed `XBPS_PASSPHRASE` fallback.**

Two problems:

1. `XBPS_PASSPHRASE` is a real environment variable read by `xbps-rindex` for its repository signing key. A user who exports it for `xbps-rindex --sign` will have nfpm silently pick it up as the passphrase for a *different* key. The failure mode is a confusing decrypt error - or a silent success with the wrong key.
2. It breaks the documented convention. Every other format reads `NFPM_PASSPHRASE` and `NFPM_<FORMAT>_PASSPHRASE` only (lines 252-271 for deb/rpm/apk, line 310 for msix). It also outranks `NFPM_PASSPHRASE`, which no other format does.

Smallest fix: delete lines 298-301, keeping `generalPassphrase` and the `NFPM_XBPS_PASSPHRASE` override below. Then drop the `XBPS_PASSPHRASE` lines from `nfpm_test.go:487-492` and from `www/content/docs/configuration.md:588`.

Test: rewrite the existing `TestOptionsFromEnvironment/xbps passphrase` subtest to assert `XBPS_PASSPHRASE` is **ignored** when `NFPM_PASSPHRASE` is set.

testdata/acceptance/xbps.dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN test -n "${scenario}"
+
+RUN mkdir -p /repo
+COPY ${oldpackage} /repo/${oldpkgfile}
+COPY ${package} /repo/${pkgfile}
+
+RUN set -eu; \
+    if [ "${scenario}" = "upgrade" ]; then \
+      xbps-rindex -a "/repo/${oldpkgfile}"; \
+      xbps-install -i -R /repo -y foo; \
```
caarlos0, line 22: **The signature sidecar is never verified by XBPS anywhere in this PR.**

`-i` is `--ignore-repository-signature`, so every acceptance scenario skips signature checking. Combined with the sidecar having no test against real tooling, the headline signing feature has no end-to-end validation - only a round-trip through nfpm's own `RSAVerifySHA256Digest`, which cannot detect a format mismatch.

I checked the format statically and it is correct: upstream `lib/verifysig.c` `rsa_verify_hash` does `RSA_verify(NID_sha256, <sha256 of the package file>, sig, siglen, rsa)` against `<file>.sig2` as raw bytes, matching `writeSignatureSidecar`. I am not claiming it is broken - I am flagging that nothing here would catch it if it were.

There is also a usability gap worth documenting: `xbps_verify_signature` returns false for an unsigned repository and pulls the public key from the repo's `index-meta.plist`. So an nfpm-produced `.sig2` is only verifiable if the user *separately* signs the repository with `xbps-rindex --sign` using the **same** key. `www/content/docs/quick-start.md` currently teaches the `-i` bypass and does not mention this prerequisite.

Two options, either is fine:
- Add a `signed` scenario that runs `xbps-rindex --sign --privkey` with the same key, then installs **without** `-i`, proving the sidecar verifies. This is the coverage the feature needs.
- Or split signing out of this PR and land the packager first.

If signing stays, please document in `configuration.md` that the repository must be signed with the same key for the sidecar to be usable.

internal/sign/rsa.go
```diff
@@ -47,45 +72,29 @@ func RSASignSHA1Digest(sha1Digest []byte, keyFile, passphrase string) ([]byte, e
...
 
-	var priv crypto.Signer
-
 	switch block.Type {
 	case PKCS1PrivkeyPreamble:
-		priv, err = x509.ParsePKCS1PrivateKey(blockData)
-		if err != nil {
-			return nil, fmt.Errorf("parse PKCS#1 private key: %w", err)
-		}
+		return x509.ParsePKCS1PrivateKey(blockData)
```
caarlos0, line 84: **Regression: the PKCS#1 error context was dropped.**

This previously wrapped the failure:

```go
priv, err = x509.ParsePKCS1PrivateKey(blockData)
if err != nil {
	return nil, fmt.Errorf("parse PKCS#1 private key: %w", err)
}
```

Now a corrupt PKCS#1 key surfaces a bare `x509: ...` with no indication of which key file or which step failed - and the PKCS#8 branch two lines below still wraps, so the two paths are now inconsistent.

Smallest fix:

```go
case PKCS1PrivkeyPreamble:
	priv, err := x509.ParsePKCS1PrivateKey(blockData)
	if err != nil {
		return nil, fmt.Errorf("parse PKCS#1 private key: %w", err)
	}
	return priv, nil
```

Test: mirror the existing `TestLoadPrivateKeyPKCS8ParseError` with a `PKCS1PrivkeyPreamble` block containing invalid DER and assert `ErrorContains(err, "parse PKCS#1 private key")`.

internal/sign/rsa.go
```diff
@@ -47,45 +72,29 @@ func RSASignSHA1Digest(sha1Digest []byte, keyFile, passphrase string) ([]byte, e
...
 		privAny, err := x509.ParsePKCS8PrivateKey(blockData)
 		if err != nil {
 			return nil, fmt.Errorf("parse PKCS#8 private key: %w", err)
 		}
-		privTmp, ok := privAny.(crypto.Signer)
+		priv, ok := privAny.(*rsa.PrivateKey)
 		if !ok {
-			return nil, fmt.Errorf("cannot sign with given private key")
+			return nil, errNoRSAKey
 		}
```
caarlos0, line 93: **Out of scope: this narrows APK signing behavior shipped in #804.**

This branch was `privAny.(crypto.Signer)`, added deliberately by 1667ef9 ("feat: support PKCS8 keys for APK signing", #804). `RSASignSHA1Digest` is called from `apk/apk.go:275`, so an ECDSA or Ed25519 PKCS#8 key that was accepted before now fails with `key is not an RSA key`.

I think the narrowing is *correct* - apk-tools verifies RSA only, so those keys produced signatures nothing could verify - and `TestLoadPrivateKeyRejectsNonRSAPKCS8` documents the reasoning well. But it is a user-visible change to a shipped format, unrelated to xbps, and it should not ride along in a packager PR where it will not get the attention it deserves.

Please restore `crypto.Signer` here and open a separate PR for the narrowing. That PR is a clean, easy yes on its own.

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+func writeSignatureSidecar(info *nfpm.Info, digest []byte) error {
+	signature, err := signPackageDigest(info, digest)
+	if err != nil {
+		return &nfpm.ErrSigningFailure{Err: err}
+	}
+	if err := os.WriteFile(info.Target+".sig2", signature, 0o644); err != nil {
+		return &nfpm.ErrSigningFailure{Err: fmt.Errorf("write signature sidecar: %w", err)}
+	}
+	return nil
+}
```
caarlos0, line 580: **A stale `.sig2` survives a failed or unsigned rebuild.**

`internal/cmd/package.go:113` does `os.Remove(target)` when `Package` fails, but nothing removes `target + ".sig2"`. Two reachable consequences when building repeatedly to the same path:

1. Sign once, then rebuild after removing `key_file` -> an orphaned `.sig2` remains next to a package it no longer matches.
2. Sign once, then a later build fails -> the package is cleaned up but its sidecar is not.

In a signed repository XBPS then fails verification against a signature for different bytes, which is a genuinely confusing failure to debug.

Smallest fix - remove any stale sidecar before writing, inside `Package`:

```go
if packageShouldBeSigned(info) {
	if err := requireSignatureTarget(info); err != nil {
		return err
	}
	if err := os.Remove(info.Target + ".sig2"); err != nil && !os.IsNotExist(err) {
		return err
	}
	...
```

That covers the failure path too, since the removal happens before any archive bytes are written.

Related, and worth a doc line rather than a code change: signing depends on `info.Target`, which only the CLI sets (`internal/cmd/package.go:109`). A library caller passing a perfectly good `io.Writer` gets a hard `target path required for signature sidecar` error. Defensible for a sidecar, but undocumented - please note it in `configuration.md`.

nfpm.go
```diff
@@ -507,6 +529,31 @@ type IPKAlternative struct {
...
+	Target   string `yaml:"target,omitempty" json:"target,omitempty" jsonschema:"title=target path"`
+}
+
+// XBPSSignature contains signing configs that are only available on XBPS packages.
+type XBPSSignature struct {
+	KeyFile       string                               `yaml:"key_file,omitempty" json:"key_file,omitempty" jsonschema:"title=RSA private key file"`
+	KeyPassphrase string                               `yaml:"-" json:"-"`
+	SignFn        func(data io.Reader) ([]byte, error) `yaml:"-" json:"-"`
+}
+
```
caarlos0, line 556: `XBPSSignature` re-declares `KeyFile`, `KeyPassphrase`, and `SignFn` - the same three fields as `PackageSignature` at line 437, with the same semantics.

`RPMSignature` (line 451) already shows the pattern:

```go
type XBPSSignature struct {
	PackageSignature `yaml:",inline" json:",inline"`
}
```

That removes three field declarations and keeps the `SignFn` doc comment in one place. It does add a `key_id` field xbps ignores - if that is the reason for the duplication, a one-line comment saying so would stop someone "fixing" it later.

Non-blocking; take it or leave a comment.

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+		}
+	}
+	for _, content := range sortedContents(info) {
+		if content.Type == files.TypeRPMGhost {
+			continue
+		}
+		if err := writeContentEntry(tw, content); err != nil {
+			_ = tw.Close()
+			_ = zw.Close()
+			return err
```
caarlos0, line 720: **Simplification:** `_ = tw.Close(); _ = zw.Close()` is repeated six times across these error paths, and the payload loop is the only place a new one could be forgotten.

Extracting the archive body lets `defer` handle it once:

```go
func writeArchive(w io.Writer, info *nfpm.Info, propsData, manifestData, installScript, removeScript []byte) (err error) {
	zw, err := zstd.NewWriter(w)
	if err != nil {
		return fmt.Errorf("xbps: create zstd writer: %w", err)
	}
	defer func() {
		if cerr := zw.Close(); cerr != nil && err == nil {
			err = fmt.Errorf("xbps: close zstd writer: %w", cerr)
		}
	}()
	tw := tar.NewWriter(zw)
	defer func() {
		if cerr := tw.Close(); cerr != nil && err == nil {
			err = fmt.Errorf("xbps: close tar writer: %w", cerr)
		}
	}()
	// ... plain sequential writes, early return on error
}
```

Behavior-preserving, and `Package` shrinks to orchestration. No new test needed - `TestPackageWritesZstdArchive` and `TestPackageReturnsLifecycleScriptReadError` already cover both paths.

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+			entry, err := fileEntry(content)
+			if err != nil {
+				return nil, err
+			}
+			configs = append(configs, entry)
+		default:
+			entry, err := fileEntry(content)
+			if err != nil {
+				return nil, err
+			}
```
caarlos0, line 424: **Simplification:** all four branches of this switch call `fileEntry(content)` and repeat the identical three-line error check. Hoisting it collapses the body to classification only:

```go
for _, content := range sortedContents(info) {
	if content.Type == files.TypeRPMGhost {
		continue
	}
	entry, err := fileEntry(content)
	if err != nil {
		return nil, err
	}
	switch {
	case content.Type == files.TypeDir, content.Type == files.TypeImplicitDir:
		dirs = append(dirs, entry)
	case content.Type == files.TypeSymlink:
		links = append(links, entry)
	case isConfigType(content.Type):
		configs = append(configs, entry)
	default:
		regular = append(regular, entry)
	}
}
```

Behavior-preserving; `fileEntry` already returns early for dirs and symlinks without opening anything. Covered by the existing `TestFilesManifestClassifiesPayloadEntries`.

xbps/xbps.go
```diff
@@ -0,0 +1,736 @@
...
+
+func configFiles(info *nfpm.Info) []string {
+	var result []string
+	for _, content := range sortedContents(info) {
+		if isConfigType(content.Type) {
+			result = append(result, files.NormalizeAbsoluteFilePath(content.Destination))
+		}
+	}
+	return result
+}
```
caarlos0, line 163: **Simplification:** both functions call `sortedContents`, which clones and sorts the slice, but neither depends on order.

- `installedSize` sums a `uint64` - order-independent by construction.
- `configFiles` hands its result straight to `stringsToPlistArray`, which sorts again via `sortedStrings`.

Iterating `info.Contents` directly drops two clones and one redundant sort, and makes the single remaining sort (in `stringsToPlistArray`) the obvious source of ordering. Output is byte-identical.

Not a performance claim - I have no benchmark and would not expect a measurable one. It is about not implying an ordering requirement that does not exist.

xbps/xbps_test.go
```diff
@@ -0,0 +1,655 @@
...
+func TestConventionalFileNameDefaultRelease(t *testing.T) {
+	info := exampleInfo()
+	info.Release = ""
+	require.Equal(t, "foo-1.2.3.beta-1.git-1_1.x86_64.xbps", Default.ConventionalFileName(info))
+}
+
+func TestConventionalFileNameNoarch(t *testing.T) {
+	info := exampleInfo()
+	info.Arch = "all"
+	require.Equal(t, "foo-1.2.3.beta-1.git-1_2.noarch.xbps", Default.ConventionalFileName(info))
```
caarlos0, line 171: These expectations encode the malformed `pkgver` from my comment on `xbps/xbps.go:65`, which is why the defect is not caught anywhere.

`foo-1.2.3.beta-1.git-1_2` is not a valid XBPS pkgver: upstream `xbps_pkg_name()` parses its name as `foo-1.2.3.beta-1.git`, not `foo`. The same broken value is asserted at lines 218, 285, 441, and 442.

Once `normalizeVersionPart` replaces hyphens instead of trimming them, these become e.g.:

```go
require.Equal(t, "foo-1.2.3.beta.1.git.1_2.x86_64.xbps", Default.ConventionalFileName(info))
```

Beyond updating the literals, please add a test that asserts the *contract*, so this cannot regress into an unparseable name again:

```go
func TestPkgverIsParseableByXBPS(t *testing.T) {
	info := exampleInfo()
	pv, err := pkgver(info)
	require.NoError(t, err)
	name, rest, ok := strings.Cut(pv, "-")
	require.True(t, ok)
	require.Equal(t, info.Name, name)
	require.NotContains(t, rest, "-", "version must not contain the name/version separator")
	require.Equal(t, 1, strings.Count(rest, "_"), "version must contain exactly one revision separator")
}
```

That test fails on the current code and passes after the fix.

caarlos0 (changes requested):
**Requesting changes.** Reviewed the full diff at `d0b8d3b` against `origin/main`. This is careful, well-structured work - the packager follows the existing first-class model closely and the hand-rolled plist marshaller is a reasonable 0-dependency choice. Two defects block it, though, and both are blessed by the current tests rather than caught by them.

**Blockers**
1. Any version containing a hyphen produces a `pkgver` that XBPS parses into the wrong package name (`xbps/xbps.go:65-84`). Reachable with a plain semver prerelease; silently produces an unusable package.
2. `arch: arm6` and `arch: arm7` cannot be packaged at all (`xbps/xbps.go:296-300`).

**Cross-cutting, no single site**

*Scope.* Two changes are not needed to add an xbps packager and change behavior for existing formats. Please split them into their own PRs so they can be judged on their own merits:
- `hasAnyArch` in `nfpm.go` (inline comment) loosens arch validation for deb/rpm/apk.
- The PKCS#8 narrowing in `internal/sign/rsa.go` (inline comment) narrows APK behavior shipped in #804.

For this PR, adding `info.XBPS.Arch != ""` to the existing condition is the whole change needed.

*Missing test coverage.* Both blockers exist because the tests assert current behavior instead of the contract:
- No test drives `Package` with `arm6`/`arm7`. `TestEnsureValidArchMappings` calls `ensureValidArch` exactly once, so it structurally cannot catch a non-idempotency bug.
- No acceptance config exercises a prerelease. `accept()` defines `SEMVER` as `v1.0.0-0.1.b1+git.abcdefgh`, but every `xbps.*.yaml` uses a plain `1.2.3`, so Docker never sees a hyphenated version. Please add one acceptance scenario that installs a prerelease build and asserts `xbps-query -p pkgname foo` returns `foo`.

*Verification gaps.* Per the requester I did not run builds or the test suite; I assume CI is green. Claims below are traced statically against the code and against upstream XBPS sources (`lib/util.c`, `lib/verifysig.c`, `bin/xbps-rindex/index-add.c`, `bin/xbps-rindex/sign.c` at `void-linux/xbps@master`). I did not run XBPS tooling myself. Notably the `.sig2` sidecar is never verified by real XBPS tooling anywhere in this PR - see the inline comment on the acceptance dockerfile.

I did confirm the sidecar format is correct: upstream `rsa_verify_hash` does `RSA_verify(NID_sha256, <sha256 of the package file>, ...)` against `<file>.sig2`, which is exactly what `RSASignSHA256Digest` produces.

Disclosure: this review was produced by an automated reviewer (GitHub Copilot CLI).

PR #1097, opened 2026-06-13
feat: add msi support

djgilcrease:
Adds an `msi` packager that produces a real Windows Installer database with the pure-Go [go-msi](https://github.com/digitalxero/go-msi) library (no WiX, no Windows needed to build or sign).

## What you get

- Contents mapped onto the standard Windows folders (`/Program Files`, `/ProgramData`, `/Windows/System32`, ...); everything else lands under the product's own folder in Program Files (or `%LOCALAPPDATA%\Programs` for `per_user` installs).
- Root metadata reused where MSI has a place for it: `description`/`homepage` become the Add/Remove Programs fields, `vendor`/`maintainer` the manufacturer, `type: license` contents the install-wizard license text, and the root `scripts` become elevated deferred custom actions with deb/rpm hook semantics.
- Under `msi:`: advertised shortcuts, Windows services, registry values, per-user installs, a canned minimal UI, and Authenticode signing with a PFX (plus RFC 3161 timestamps).
- Every release replaces older installs of the same upgrade code in place, and a downgrade is refused (`allow_downgrades` / `allow_same_version` are available). The upgrade code and component GUIDs are derived from identities that survive releases; the product code follows the full version.

## Verification

- Unit tests build real packages, run go-msi's ICE validator, and assert on the decoded tables (go-msi v0.5.0 exports a read-only reader for this). Golden GUIDs pin the derivation contract.
- `TestMSIStructure` (Linux) builds amd64/arm64 packages from a real config.
- On a Windows runner: install, run the app, check the service/registry/shortcut, upgrade to the next version in place (exactly one product registered), uninstall (everything gone), and prove a failing `postinstall` rolls the install back. A second job validates and installs a package that nfpm signed on Linux.

## Commits

1. `chore(deps)`: go-msi v0.5.0, go-pkcs12 v0.7.3
2. `feat: add msi support`
3. `ci`: the signed-package validation job

The MSIX behavior changes that used to live here are now #1135.


djgilcrease:
@copilot resolve the merge conflicts in this pull request

djgilcrease:
@caarlos0 feedback / thoughts?

caarlos0:
@djgilcrease sorry was short on time this past few weeks, gonna review this now :D

www/go.mod
```diff
@@ -1,5 +1,3 @@
 module github.com/goreleaser/nfpm/hugo-docs
 
 go 1.21
-
-require github.com/imfing/hextra v0.12.3 // indirect
```
caarlos0, line 5: this and www/go.mod.sum need to be reverted

caarlos0 (changes requested):
one small thing


other than that, i almost feel like the windows packaging should live in a different project, as nfpm was mostly thought out to handle Linux only... e.g. it doesn't reuse most of the fields from the root structure, and mainly use the its own fields. Same for msix that is already merged.

Do you have any opinions on this? maybe a `wpm` project (windows package manager) or something like that

djgilcrease:
> one small thing
> 
> other than that, i almost feel like the windows packaging should live in a different project, as nfpm was mostly thought out to handle Linux only... e.g. it doesn't reuse most of the fields from the root structure, and mainly use the its own fields. Same for msix that is already merged.
> 
> Do you have any opinions on this? maybe a `wpm` project (windows package manager) or something like that

Let me see if I can refactor them to be more shared, if not, I can see splitting it to a wpm, that would focus on msi/msix/nuget/winget/etc

www/go.mod
djgilcrease: fixed

djgilcrease:
Ok, I cleaned it up a bit to default to using the existing fields, including the license type for content instead of a special msi.license field.

If you still feel this should be split, I am fine with that and will work on creating that.

msi/msi.go
```diff
@@ -0,0 +1,663 @@
...
+
+	// ProductCode must always be present. When omitted we derive a stable GUID
+	// from the product name (kept constant across versions so it does not change
+	// on every version bump).
+	// go-msi requires braced uppercase GUIDs; validation accepts either case, so
+	// normalize the user-provided value here.
+	productCode := strings.ToUpper(info.MSI.ProductCode)
+	if productCode == "" {
+		productCode = deriveGUID("product|" + info.MSI.ProductName)
+	}
```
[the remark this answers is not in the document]
djgilcrease: fixed

msi/msi.go
```diff
@@ -0,0 +1,663 @@
...
+
+	if err := validate(info); err != nil {
+		return err
+	}
+
+	b := msi.NewPackage().
+		WithProductName(info.MSI.ProductName).
+		WithManufacturer(info.MSI.Manufacturer).
+		WithVersion(convertToMSIVersion(info.Version)).
+		WithAllUsers(*info.MSI.AllUsers)
```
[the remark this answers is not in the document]
djgilcrease: fixed

msi/msi.go
```diff
@@ -0,0 +1,663 @@
...
+	return nil
+}
+
+// nolint: gochecknoglobals
+var startTypes = map[string]msi.ServiceStartType{
+	"auto":     msi.ServiceStartAuto,
+	"demand":   msi.ServiceStartDemand,
+	"disabled": msi.ServiceStartDisabled,
+	"boot":     msi.ServiceStartBoot,
+	"system":   msi.ServiceStartSystem,
```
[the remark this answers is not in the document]
djgilcrease: fixed

msi/msi.go
```diff
@@ -0,0 +1,663 @@
...
+	parts := strings.SplitN(version, ".", 4)
+	result := make([]string, 3)
+	for i := range 3 {
+		result[i] = "0"
+		if i < len(parts) {
+			if n, err := strconv.Atoi(parts[i]); err == nil {
+				if n > 65535 {
+					n = 65535
+				}
+				result[i] = strconv.Itoa(n)
```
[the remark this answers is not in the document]
djgilcrease: fixed

djgilcrease:
Added signing validation based on the fix in digitalxero/go-msix#4

msi/msi.go
```diff
@@ -0,0 +1,763 @@
...
+			dirID := makeID("d", accum)
+			if !createdDirs[dirID] {
+				b.Directory(parentID).Subdirectory(dirID, seg)
+				createdDirs[dirID] = true
+			}
+			parentID = dirID
+		}
+
+		compID := makeID("c", dest)
+		comp := b.Directory(parentID).Component(compID).WithGUID("")
```
caarlos0: **Blocking — default config loses user files on uninstall.**

`WithGUID("")` makes go-msi derive the component GUID, and its seed includes the ProductCode:

```go
// go-msi compile.go:333
seed := "component|" + p.productCode + "|" + e.dirID + "|" + cid
```

`deriveProductCode` (msi.go:754) includes `msiVersion(info)`, so **the same file at the same path gets a different component GUID in every release**. That is exactly what the Windows Installer component rules forbid.

It only bites on defaults because `msi.upgrade.enabled` is `false` by default (msi.go:237), so no `Upgrade` table and no `RemoveExistingProducts` is emitted. With `enabled: true`, go-msi's `RemoveAfter` defaults to `InstallInitialize`, the old product is removed before the new files land, and the divergent GUIDs are harmless.

**Trigger** — default config, nothing but `name`/`version`/`contents`:
1. Package and install v1.0.0 → ProductCode P1, component GUIDs A.
2. Package and install v1.1.0 → ProductCode P2, component GUIDs B. Without an `Upgrade` table the UpgradeCode is an inert Property row, so msiexec treats this as an unrelated product: two entries in Apps & Features, v2's files overwriting v1's in place.
3. Uninstall v1.0.0 → set A's refcounts hit zero (B does not share them) → **the files and registry keys v1.1.0 is still using are deleted.**

Every other nfpm packager upgrades in place by default; this one silently breaks the installed product.

**Smallest fix** — seed the component GUID from the *upgrade* code, which is stable across releases:

```go
comp := b.Directory(parentID).Component(compID).
    WithGUID(deriveGUID("component|" + upgradeCode + "|" + dest))
```

Same at msi.go:504 for registry components. This is correct under both `upgrade.enabled` settings, so it fixes the bug independently of the default. I would also flip the default (see my comment on nfpm.go:608).

**Test** — build the same `info` twice with only `Version` differing and assert the component GUID for a given destination is identical. `TestDerivedProductCode` (msi_test.go:285) asserts byte-identical output for *identical* input, which is the opposite invariant, so nothing today catches this. An acceptance step that installs v1, installs v2, uninstalls v1, and asserts the v2 executable still exists would catch it end to end.

msi/scripts.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return "", err
+	}
+	run, ok := scriptExtensions[strings.ToLower(filepath.Ext(path))]
+	if !ok {
+		return "", fmt.Errorf("%q: msi scripts must be .ps1, .bat, or .cmd", path)
+	}
+
+	bootstrap := fmt.Sprintf(
+		"$f = Join-Path $env:TEMP 'nfpm_%s%s'\n"+
```
caarlos0: **Security — local privilege escalation via a predictable temp file.**

The bootstrap writes the extracted script to a **fixed, predictable** path, and the custom action is `Deferred().NoImpersonate()` (scripts.go:73), so it runs as SYSTEM. SYSTEM's `%TEMP%` is `C:\Windows\Temp` — `install-msi.ps1:8` relies on exactly that.

`C:\Windows\Temp` grants `BUILTIN\Users` create-file rights and gives `CREATOR OWNER` full control of what they create. So an unprivileged local user can pre-create `nfpm_NfpmPreInstall.ps1`, own it, and keep a writable handle. When an admin later runs the installer, SYSTEM truncates and writes the real script, then executes it — leaving a TOCTOU window the attacker can win (trivially with an oplock). Result: arbitrary code as SYSTEM. The attacker can also ACL the file to deny SYSTEM and turn every install into a failure. This is CWE-377, and it fires on any machine where a non-admin has a local account.

**Smallest fix** — use an unpredictable name:

```go
"$f = Join-Path $env:TEMP ('nfpm_%s_' + [IO.Path]::GetRandomFileName() + '%s')\n"+
```

(`GetRandomFileName` is CSPRNG-backed; keep the extension suffix so the interpreter dispatch still works.) Creating a fresh per-action directory with a restrictive ACL and writing inside it is stronger still, if you want belt and braces.

**Test** — assert in `TestScriptCommandLineEmbedsScript` that the decoded bootstrap does *not* contain a literal `nfpm_NfpmPreInstall.ps1`, i.e. that the filename is generated at run time rather than baked in.

msi/msi.go
```diff
@@ -0,0 +1,763 @@
...
+	// UpgradeCode must stay stable across releases of the same product so the
+	// Upgrade table can find older installs; namespace it by manufacturer,
+	// product name, and architecture (per-arch products, no cross-arch
+	// upgrades).
+	upgradeCode := strings.ToUpper(info.MSI.UpgradeCode)
+	if upgradeCode == "" {
+		upgradeCode = deriveUpgradeCode(info)
+	}
+	b = b.WithUpgradeCode(upgradeCode)
+	for k, v := range info.MSI.Properties {
```
caarlos0, line 193: **`msi.properties` can emit a duplicate `Property` primary key.**

Every user key is forwarded to `WithProperty` unfiltered, but go-msi emits the identity properties in a *separate* `WithProperties` call (`compile.go:155-185`: `ProductName`, `ProductVersion`, `Manufacturer`, `ProductCode`, `ALLUSERS`, `UpgradeCode`) and `addRow` does not deduplicate. go-msi guards only two keys — `SecureCustomProperties` (compile.go:191) and `ProductLanguage` (compile.go:169) — which is itself evidence that duplicates are the failure mode here.

**Trigger:** `msi.properties: {ProductVersion: "9.9"}` → two `ProductVersion` rows → a database with a duplicate PK and undefined resolution. The docs actively invite users into `msi.properties` ("Setting ARPCOMMENTS or ARPURLINFOABOUT here overrides…"), so this is reachable from documented config.

**Smallest fix** — reject the reserved keys in `validate()`:

```go
for k := range info.MSI.Properties {
    if reservedProperties[k] {
        return fmt.Errorf("package msi.properties[%q] is set by nfpm; use the dedicated field instead", k)
    }
}
```

with `reservedProperties` covering `ProductName`, `ProductVersion`, `Manufacturer`, `ProductCode`, `UpgradeCode`, and `ALLUSERS`.

**Test** — a table case in `msi_test.go` asserting `Package` fails for each reserved key.

msi/msi.go
```diff
@@ -0,0 +1,763 @@
...
+	{"windows/fonts", "FontsFolder", "FontsFolder"},
+	{"windows", "WindowsFolder", "WindowsFolder"},
+	{"appdata/local", "LocalAppDataFolder", "LocalAppDataFolder"},
+	{"appdata/roaming", "AppDataFolder", "AppDataFolder"},
+}
+
+// systemDirs are directories whose components should be marked Permanent to
+// satisfy ICE09.
+// nolint: gochecknoglobals
+var systemDirs = map[string]bool{
```
caarlos0, line 786: **Files installed into system directories are never removed, and nothing says so.**

Marking these components `msidbComponentAttributesPermanent` does satisfy ICE09, but Permanent means Windows Installer never removes them. A user who writes `dst: /Windows/System32/foo.dll` gets a file that outlives every uninstall, with no warning at package time and no mention in `configuration.md`.

The ICE09 rationale in the comment is accurate; the user-visible consequence is what is missing.

**Smallest fix** — document it in the `msi:` destination-mapping section of `www/content/docs/configuration.md`, and `log.Printf` a warning when a content lands in a `systemDirs` root, matching how the symlink skip at msi.go:333 already warns.

**Test** — extend `TestPackageSystemDir` (msi_test.go:753) to capture the log output and assert the warning fires; it currently only asserts the build succeeds.

nfpm.go
```diff
@@ -573,6 +584,70 @@ type MSIXSignature struct {
...
+	Upgrade      MSIUpgrade        `yaml:"upgrade,omitempty" json:"upgrade,omitempty" jsonschema:"title=major upgrade behavior"`
+	Shortcuts    []MSIShortcut     `yaml:"shortcuts,omitempty" json:"shortcuts,omitempty" jsonschema:"title=shortcuts to create"`
+	Services     []MSIService      `yaml:"services,omitempty" json:"services,omitempty" jsonschema:"title=windows services to install"`
+	Registry     []MSIRegistry     `yaml:"registry,omitempty" json:"registry,omitempty" jsonschema:"title=registry entries to create"`
+	Signature    MSISignature      `yaml:"signature,omitempty" json:"signature,omitempty" jsonschema:"title=msi signature"`
+}
+
+// MSIUpgrade contains major-upgrade configuration for MSI packages.
+type MSIUpgrade struct {
+	Enabled               bool   `yaml:"enabled,omitempty" json:"enabled,omitempty" jsonschema:"title=enable WiX-style major upgrade handling"`
```
caarlos0: **`upgrade.enabled` is the wrong shape, and the wrong default.**

Defaulting to `false` is what makes the component-GUID bug at msi/msi.go:416 fire on the default path. Beyond that: what is the concrete user who wants a package that installs *alongside* its own previous version? nFPM already emits a stable UpgradeCode and a per-release ProductCode unconditionally — everything an upgrade needs except the one table this flag gates.

I would drop `enabled` entirely and keep `MSIUpgrade` as just `downgrade_error_message`, always calling `b.MajorUpgrade()`. That removes a knob, removes the footgun, and matches what every other nfpm packager does by default. If you would rather keep the escape hatch, invert it so the safe path is the default — but I would want to see the user who asks for it first.

The docs already promise behavior this default does not deliver: the `upgrade_code` block says the derived code "stays constant across releases and upgrades work", which is only true with `enabled: true`. Please also warn there that *pinning* `product_code` makes `msiexec /i` of v2 over v1 fail with 1638.

While here: `AllUsers *bool` (nfpm.go:596) is the first tri-state pointer in this file, used to express two states. `PerUser bool` defaulting to false gives identical behavior with no pointer and no `SetPackagerDefaults` allocation.

**Test** — assert an `Upgrade` table is emitted for a default-config package; `TestMajorUpgrade` (msi_test.go:407) only asserts ICE-cleanliness with the flag on, and nothing covers the default at all.

msi/msi_test.go
```diff
@@ -0,0 +1,998 @@
...
+	info.MSI.Registry = []nfpm.MSIRegistry{
+		{Root: "HKXX", Key: `Software\TestCo`, Name: "x", Value: "y"},
+	}
+	var buf bytes.Buffer
+	err := msi.Default.Package(info, &buf)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "root")
+}
+
+func TestMajorUpgrade(t *testing.T) {
```
caarlos0: This passes if `MajorUpgrade()` is never called — it only asserts the build succeeds and is ICE-clean, and an MSI with no `Upgrade` table is perfectly ICE-clean. The same is true of `TestShortcut` (:302), `TestService` (:325), `TestRegistry` (:387), and `TestPackageSystemDir` (:753): each would still pass if the corresponding `add*` function became a no-op, so none of them can catch a regression in the feature it is named for.

go-msi exports no table reader, so the pragmatic assertion is `bytes.Contains` on the returned bytes — the technique `TestManufacturerFallback` (:201) already uses and `TestMSIStructure` uses for the platform string. Asserting that the shortcut name, service name, registry key, and the `downgrade_error_message` string are present in the built package is weak, but it is the difference between catching a dropped feature and catching nothing.

The assertion that actually matters is the one described in my comment on msi/msi.go:416: build twice with only `Version` differing and compare component GUIDs.

msi/msi.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+	if systemDirs[rootID] {
+		attrs |= msidbComponentAttributesPermanent
+	}
+	return attrs
+}
+
+// normalizeDest converts a destination path to a forward-slash relative path
+// with any drive letter and leading slashes removed.
+func normalizeDest(p string) string {
```
caarlos0, line 839: Two destination inputs produce a bad result rather than a clear error:

1. `dst: "C:/"` (nfpm normalizes it to `/C:`) → `normalizeDest` returns `""` → `mapDestination` returns `rel=""` → `fileName` is `""` → `WithFile("", src)`, which fails deep inside go-msi's long-name validation. The user gets an opaque library error instead of "destination must not be empty".
2. Drive letters are discarded, so `C:/Program Files/App/x.txt` and `/Program Files/App/x.txt` collapse to the same `dest`, hence the same `compID` and the same generated File ID — a duplicate `File` primary key. `files.PrepareForPackager` cannot catch it because the destinations differ before normalization.

**Smallest fix** — in `validate()`, reject a content whose `normalizeDest` is empty, and reject two contents that normalize to the same path.

**Test** — table cases in `coverage_test.go:61` for `"C:/"`, `"C:"`, and `"/C:"` asserting a named validation error, plus one `Package` case with the two colliding destinations above.

msi/scripts.go
```diff
@@ -0,0 +1,150 @@
...
+	"unicode/utf16"
+
+	"github.com/goreleaser/nfpm/v2"
+	"go.digitalxero.dev/go-msi"
+)
+
+// maxScriptSize bounds the size of an embedded maintainer script. Scripts are
+// double-base64-encoded into the custom action command line (~3.6x expansion),
+// which Windows caps at 32,767 characters.
+const maxScriptSize = 8 * 1024
```
caarlos0, line 31: The constant is fine; the comment's arithmetic is not, and the boundary is untested.

Reproducing the `Sprintf` + UTF-16LE + base64 chain gives a `CustomAction.Target` of 875 chars for a 0-byte script and 30 003 chars for 8 192 bytes; the largest script that stays under 32 767 is 8 970 bytes. So the factor is 3.66x at the limit, not "~3.6x" of the script alone, and the ~9% headroom reads accidental rather than chosen.

Worth noting in the comment: `CustomAction.Target` is declared 255 chars wide in the MSI schema (go-msi `table_catalog.go:650`). go-msi's ICE03 treats `Formatted` as free-form and never flags the overflow, and msiexec accepts it — your Windows acceptance job proves that for the ~1.1k-char targets it builds. But real ICE03 (`msival2`, `wix light`) does report it, which will surprise anyone running this MSI through a corporate validation gate. One sentence saying that 255 is unattainable here and deliberately exceeded would save the next reader the investigation.

**Test** — `TestScriptTooLarge` (msi_test.go:546) covers the reject side only. Add the accept-side boundary: generate a `maxScriptSize`-byte script and assert `len(scriptCommandLine(...)) <= 32767`. That is what catches drift if the bootstrap template ever grows.

msi/msi.go
```diff
@@ -0,0 +1,763 @@
...
+// msiVersionMaxima are the per-field maxima Windows Installer enforces on the
+// ProductVersion property: Major.Minor.Build, capped at 255.255.65535. A value
+// outside them makes the package uninstallable.
+// nolint: gochecknoglobals
+var msiVersionMaxima = [3]int{255, 255, 65535}
+
+// convertToMSIVersion converts a semver-style version to MSI's
+// Major.Minor.Build format. Each field is numeric and clamped to the maximum
+// Windows Installer accepts for it.
+func convertToMSIVersion(version string) string {
```
caarlos0: `convertToMSIVersion` has no production caller — `msiVersionClamped` calls `convertToMSIVersionClamped` directly, and the only references are `coverage_test.go:242` and `:256`. Delete it and have those two tests call `convertToMSIVersionClamped`, discarding the second return.

msix/msix.go
```diff
@@ -66,6 +67,47 @@ func (*MSIX) ConventionalExtension() string {
 	return ".msix"
 }
 
+// vendorOrMaintainer returns info.Vendor, falling back to the name part of
+// info.Maintainer ("Jane Doe <jane@example.com>" -> "Jane Doe").
+func vendorOrMaintainer(info *nfpm.Info) string {
```
caarlos0: `vendorOrMaintainer` is byte-identical to `msi/msi.go:107`. Both copies are introduced by this PR, so this is the moment to avoid the duplication — one copy in an internal package, or none here at all if the MSIX changes move to their own PR as suggested in the review body.

The MSIX behavior changes anchored around here (publisher defaulted from the signing certificate at :174, `publisher_display_name` falling back to vendor/maintainer at :248, the scripts warning at :167, and `publisher` dropping out of `required` in `schema.json`) all look correct and are well covered by `msix_test.go` and `msix/coverage_test.go`. My only objection is that they change a shipped packager's behavior inside a PR titled `feat: add msi support`, where an MSIX user would never look for them.

testdata/acceptance/install-msi.ps1
```diff
@@ -0,0 +1,62 @@
...
+}
+Write-Host "Package uninstalled successfully"
+
+foreach ($hook in @('preremove', 'postremove')) {
+    if (-not (Test-Path $markers[$hook])) {
+        Write-Error "$hook script did not run (marker $($markers[$hook]) missing)"
+        exit 1
+    }
+}
+Write-Host "Remove scripts ran correctly"
```
caarlos0: This script is the strongest verification in the PR — real `msiexec`, all four hook markers, the installed binary actually executed — but every assertion is on the happy path.

Two additions would cover the claims the docs make and this PR's riskiest behavior:

1. **Rollback on script failure.** `configuration.md` states "a non-zero exit rolls back the transaction (including on uninstall)". Nothing tests it. A second config whose `postinstall` is `exit 1`, asserting `msiexec /i` returns non-zero *and* that the product is absent afterwards, would prove the deferred / no-`msidbCustomActionTypeContinue` reasoning in `scripts.go:69-73` end to end.
2. **v1 to v2 upgrade.** Install v1.0.0, install v1.0.1, assert exactly one entry in Apps & Features, then uninstall and assert the install directory is gone. This is the sequence that surfaces the component-GUID bug in my comment on `msi/msi.go:416`, and it is cheap here because the runner is already installing and removing packages.

One behavior worth documenting rather than testing: because the remove hooks abort the transaction, a failing `postremove` leaves the product installed and un-removable until the script is fixed, and `HideTarget()` means the verbose log shows nothing useful for diagnosing it.

caarlos0 (changes requested):
Automated review by GitHub Copilot CLI (a bot), acting for @caarlos0. Two independent checks (maintainer + adversarial) were run and their claims re-verified against the `go-msi@v0.4.0` source.

This is good work — a real Windows Installer database, ICE validation in the unit tests, and a genuine `msiexec` install/uninstall acceptance job on a Windows runner. That last part is more verification than most packagers here shipped with. The blocker below is a default-path data-loss bug, not a design objection.

## Verdict

One blocking finding (inline, `msi/msi.go:416`): with the **default** configuration, every release gets new component GUIDs *and* no `Upgrade` table, so installing v1.1.0 over v1.0.0 leaves two products, and uninstalling v1.0.0 deletes the files v1.1.0 needs. Both independent checks found this separately; I confirmed the mechanism in `go-msi/compile.go:333`.

One security finding (inline, `msi/scripts.go:98`) and nine smaller items follow inline.

## Cross-cutting: scope

The MSIX changes do not belong in a commit titled `feat: add msi support`. `nfpm.go:289` makes `$NFPM_PASSPHRASE` apply to MSIX signing for the first time, `msix/msix.go:174-186` makes `msix.publisher` optional, and `schema.json` drops `publisher` from the MSIX `required` set. Each change looks correct and is tested — but they are user-visible changes to a shipped packager. Please split them into their own PR so they can be released and reverted independently. The `chore(deps)` bumps belong in a third commit.

## Cross-cutting: test coverage

`packageAndValidate` building a real MSI and running the full ICE validator is the right shape, and `coverage_test.go` is genuine table-driven unit testing despite the name (please rename it or merge it into `msi_test.go` — the name suggests coverage padding, which it is not).

The gap is that most feature tests assert only "builds and is ICE-clean". `TestShortcut`, `TestService`, `TestRegistry`, `TestMajorUpgrade`, and `TestPackageSystemDir` would all still pass if the corresponding `add*` function became a no-op. These contracts have no assertion at all:

- component GUID stability across versions (the blocking finding above)
- custom action scheduling: anchors, before/after, per-hook conditions
- `install_dir` reaching INSTALLFOLDER's DefaultDir
- `all_users: false`
- `minimal_ui`
- the `upgrade.enabled: false` default
- the accept side of `maxScriptSize`

`go-msi` exports no table reader, so `bytes.Contains` on the raw CFB (already used by `TestManufacturerFallback`) is the pragmatic assertion here — it is still far better than asserting nothing.

The acceptance script (`install-msi.ps1`) covers install through uninstall well, but only the happy path. The docs claim "a non-zero exit rolls back the transaction (including on uninstall)"; nothing tests it. A v1 to v2 upgrade step would also have caught the blocking finding.

## Verification gaps to disclose

Per instruction I did not build or run tests; nothing here comes from an executed test. Reasoning is from reading this diff and the `go-msi@v0.4.0` source over the GitHub API (the module cache was not writable in this sandbox). Specifically unverified: the runtime behavior of a `CustomAction.Target` longer than the declared 255-char schema width on Windows versions other than the CI runner, and the exact failure mode of the destination collision noted at `msi/msi.go:605`.

msi/msi.go
djgilcrease: Fixed at both layers. go-msi v0.5.0 now seeds derived component GUIDs from the UpgradeCode (falling back to the ProductCode) plus the platform, and nfpm passes explicit GUIDs anyway: `component|<upgradeCode>|<arch>|<destination>` (and the registry analogue), so nfpm owns its compatibility contract regardless of the library default. The platform/arch is in the seed because a pinned `upgrade_code` shared by x86 and x64 builds must not give distinct components the same GUID.

Tests: `TestComponentGUIDsStableAcrossReleases` builds 1.0.0 and 1.1.0 and asserts identical component GUIDs per component (and distinct product codes); `TestComponentGUIDsDifferPerArch` covers the pinned-upgrade-code case; `TestDerivedCodes` pins golden values for the upgrade code and a component GUID so a seed change cannot slip through. The Windows acceptance job now installs v1.0.0, installs v1.0.1 over it, asserts exactly one product is registered at the new version with the app still running, and uninstalls.

msi/scripts.go
djgilcrease: Fixed. The bootstrap now builds the name from `[IO.Path]::GetRandomFileName()` and opens it with `[IO.File]::Open($f, 'CreateNew')`, so a pre-created file fails the action (and rolls the install back) instead of being truncated and executed. `TestScriptCommandLineEmbedsScript` asserts the literal name is gone and both `GetRandomFileName` and `CreateNew` are present, and `TestScriptCommandLineFitsCommandLine` guards the accept side of the size limit (the larger bootstrap still leaves headroom under 32,767).

msi/msi.go
djgilcrease: Fixed as suggested, with one correction to the consequence: go-msi's serializer already rejected the duplicate primary key, so this was an opaque build error (`serialize table Property: duplicate primary key`) rather than an undefined database. `validate()` now rejects `ProductName`, `ProductVersion`, `Manufacturer`, `ProductCode`, `UpgradeCode`, `ALLUSERS` and `ProductLanguage` with a pointer to the right field (`TestReservedProperties`), and go-msi v0.5.0 fails fast with the key name too.

msi/msi.go
djgilcrease: Documented in the `msi:` destination-mapping section (files under `/Windows` are permanent and outlive every uninstall), and `addContents` now logs a warning once per system folder. `TestPackageSystemDir` asserts the Permanent bit on the component and that the warning fires exactly once.

nfpm.go
djgilcrease: Went with dropping it: `MSIUpgrade` is now `downgrade_error_message` plus `allow_downgrades` and `allow_same_version` passthroughs, and `MajorUpgrade()` is always called. `AllUsers *bool` became `PerUser bool` (a per-user package also gets its own validation and installs under `%LOCALAPPDATA%\Programs`). The docs now say why pinning `product_code` breaks upgrades (1638). `TestMajorUpgradeDefault` asserts the two Upgrade rows, `RemoveExistingProducts` before `InstallFiles`, and the launch condition for a default config; `TestMajorUpgradeOptions` covers the two new knobs.

msi/msi_test.go
djgilcrease: go-msi v0.5.0 exports a read-only `Open`/`Database` reader, so the tests now assert on the decoded tables instead of `bytes.Contains`: shortcut name/directory/target, ServiceInstall and ServiceControl events, Registry rows plus the component KeyPath and attributes, Upgrade rows, custom action type/sequence/conditions, INSTALLFOLDER's parent, and so on. The version-drift assertion is `TestComponentGUIDsStableAcrossReleases`.

msi/msi.go
djgilcrease: Both fixed in `validate()`, with one refinement on the first: `dst: "C:/"` actually works (the glob treats the trailing slash as a directory and the file lands under INSTALLFOLDER); it is `dst: "C:"` and `"/C:"` that normalize to an empty name. Those are now rejected with "destination ... is empty once the drive letter is removed", and two contents that normalize to the same Windows path are rejected with both spellings named. `TestDestinations` covers all four cases.

msi/scripts.go
djgilcrease: Comment rewritten with the real numbers (fixed ~1,200-character bootstrap overhead after the random-name change, 3.56 characters per script byte, and the 255-character schema width being deliberately exceeded and why). `TestScriptCommandLineFitsCommandLine` generates a `maxScriptSize` script for every hook and asserts the target stays under 32,767; `TestScriptSizeLimit` pins the 8192/8193 boundary end to end.

msi/msi.go
djgilcrease: Deleted; the tests call `convertToMSIVersionClamped` directly.

msix/msix.go
djgilcrease: The MSIX changes moved to #1135, and the helper now lives in `internal/maintainer` (used by both packagers; both branches add the identical file so they merge in either order).

testdata/acceptance/install-msi.ps1
djgilcrease: Both added. `install-msi.ps1` now takes `-Upgrade` and `-Rollback` packages: after the v1.0.0 checks it installs v1.0.1 over it and asserts exactly one product is registered at the new version (files, shortcut, service and registry still in place, remove hooks not fired), uninstalls and asserts everything is gone, then installs a package whose `postinstall` is `exit 1` and asserts msiexec fails and no product/directory is left behind. The uninstall is logged too. The failing-postremove behavior is documented in the scripts section.

djgilcrease:
Thanks for the thorough review. Everything above is addressed (replies inline), the MSIX changes are split out to #1135, and the branch is rebased into the three commits you asked for (deps / feat / ci).

While going through it I also ran a full pass over the implementation and fixed a few things the review did not raise. The root-cause ones are in **go-msi v0.5.0**, which this PR now pins:

- Registry components carried a KeyPath without the `RegistryKeyPath` attribute, so Windows Installer would have looked the value up in the File table. go-msi now sets the bit for `AsKeyPath` registry values, nfpm sets it explicitly too, and go-msi's ICE18/ICE92 read the right column (they were reading `Condition`), so the validator would actually have caught it.
- ICE80 now also flags a 64-bit component under a 32-bit-only folder; nfpm derives the 64-bit bit from the folder, so `/Program Files (x86)` and `/Windows/SysWOW64` content in an x64 package is 32-bit.
- Sibling directories with a common 6-character prefix got the same 8.3 short name; `Shortcut.Name` was emitted without the `short|long` form.
- go-msi exports a read-only table reader (`msi.Open`), which is what the tests now assert against.

nfpm-side, beyond the inline items:

- The install folder now lives under Program Files (per-machine) or `%LOCALAPPDATA%\Programs` (per-user) instead of the drive root, so Linux-style destinations like `/usr/bin/app` end up in a sane place. A per-user package is validated against Program Files / Windows / HKLM / services.
- A version that does not fit `255.255.65535` is an error instead of a clamp: two calendar versions clamped to the same value produced byte-identical packages with the same ProductCode. The product code is also seeded from the full version, so a pre-release and its release get distinct codes.
- Services are always stopped before their files are touched and stopped+deleted on uninstall (the `stop` knob is gone); the default left an orphaned service registration.
- Shortcut `directory` is validated against the standard folders; identifier hash collisions fail the build instead of merging two files; `msi.properties` values and the other free-form MSI strings are env-expanded like `deb.fields`.
- Docs cover install location, permanent system files, per-user rules, the version rule, the pre-release case, service semantics, the failing-postremove and Group Policy execution-policy caveats, and `arm` being Windows RT only.

CI: the Windows job now installs, upgrades in place, uninstalls, and proves a failing `postinstall` rolls back; the signed job installs a package signed on Linux.

PR #1113, opened 2026-07-11
fix(apk): normalize file modes before writing tar headers

### LINKED ISSUE #1112: apk: any `type: tree` entry produces a package apk-tools rejects ("package file format error") - walked dir mode overflows tar octal field
### What happened?

Building an **apk** package with any `contents` entry of `type: tree` produces a package that apk-tools rejects at install time:

```
ERROR: repro-tree-1.0.0: package file format error
```

Tested on Alpine 3.20 (apk-tools 2.14.4). A tree containing a single 6-byte text file is enough to trigger it. The **same file shipped as a plain `src:`/`dst:` entry installs fine**, and deb/rpm/archlinux handle the identical tree entry correctly.

I expected the tree-built apk to install the same as the per-file one.

### Root cause

This is the apk sibling of #1059 (rpm CPIO `c_mode` corruption). #1060 fixed the mode conversion in `rpm/rpm.go` only; the apk path still writes the raw walked mode.

- The tree walk stores Go's `fs.FileMode` — including the type bits — into `FileInfo.Mode` (`files/files.go`), so a walked directory gets `0x800001ED` (`fs.ModeDir | 0755`).
- `apk/apk.go` writes that straight into the tar header for `TypeDir`/`TypeImplicitDir` entries: `Mode: int64(file.FileInfo.Mode)`.
- `0x800001ED` doesn't fit the tar header's 7-digit octal mode field, so Go's `archive/tar` silently switches **that one header** to GNU format with base-256 (binary) numeric encoding. apk-tools only accepts USTAR/PAX headers with octal fields → "package file format error".

Raw header of the walked directory inside the data segment (`repro-tree_1.0.0_x86_64.apk`):

```
name='usr/lib/repro/'  magic=b'ustar  \x00' (GNU)  mode bytes = 80 00 00 00 80 00 01 ed  (base-256 0x800001ED)
```

versus the same directory in the plain-entry package:

```
name='usr/lib/repro/'  magic=b'ustar\x00' '00' (PAX)  mode = 0000755 (octal ASCII)
```

`deb/deb.go` already masks to permission bits (`fm & 0o7777`) — doing the same in `apk/apk.go` (or centrally where the tree walk fills `FileInfo.Mode`, which would also cover `ipk/ipk.go`, flagged in #1059 too) fixes it.

`nfpm-tree.yaml`:

```yaml
name: repro-tree
arch: amd64
platform: linux
version: 1.0.0
maintainer: repro <repro@example.com>
description: minimal tree repro
contents:
  - src: /repro/payload
    dst: /usr/lib/repro
    type: tree
```

Run on `alpine:3.20` (e.g. `docker run --rm -v "$PWD:/repro" alpine:3.20 sh /repro/repro.sh`):

```sh
#!/bin/sh
set -eu
apk add --no-cache curl tar >/dev/null
curl -fsSL https://github.com/goreleaser/nfpm/releases/download/v2.47.0/nfpm_2.47.0_Linux_x86_64.tar.gz \
  | tar -xz -C /usr/local/bin nfpm

mkdir -p /repro/payload /out
printf 'hello\n' > /repro/payload/hello.txt   # 6 bytes

nfpm package -f /repro/nfpm-tree.yaml -p apk -t /out/
apk add --allow-untrusted /out/repro-tree_1.0.0_x86_64.apk
# => ERROR: repro-tree-1.0.0: package file format error
```

Replacing the tree entry with the expanded equivalent installs cleanly:

```yaml
contents:
  - src: /repro/payload/hello.txt
    dst: /usr/lib/repro/hello.txt
```

(That per-file expansion is the workaround I'm shipping with in the meantime.)

### nfpm version

```bash
GitVersion:    2.47.0
GitCommit:     40c7c8f7376400e9464d04a2a099045cf5598f8e
GitTreeState:  false
BuildDate:     2026-06-20T23:17:25Z
BuiltBy:       goreleaser
GoVersion:     go1.26.4
Compiler:      gc
ModuleSum:     h1:0bioJAjWaMPntgDqynP4ze0Wt4zYqYSFJ5/BBy9XIGI=
Platform:      linux/amd64
```

### Search

- [x] I did search for other open and closed issues before opening this. (#805 was the same symptom, fixed in 2.38.0; #1059/#1060 fixed the same mode-bit leak for rpm only.)

### Code of Conduct

### Additional context

apk-tools is the strictest consumer here, but any tar-based packager writing `FileInfo.Mode` unmasked will emit GNU base-256 headers whenever the walked mode exceeds `0o7777777` — for directories that's always, since `fs.ModeDir` is bit 31.

ychampion:
## Summary

- normalize Go `fs.FileMode` values before writing APK tar headers
- preserve configured and Go-encoded setuid, setgid, and sticky bits
- cover walked tree directories and regular-file special modes

## Why

`type: tree` directories carry Go's `ModeDir` flag in `FileInfo.Mode`. Writing that raw value overflows tar's octal mode field, switches the header to GNU/base-256 encoding, and makes apk-tools reject the package.

Fixes #1112

## Validation

- `GOTOOLCHAIN=go1.26.4 go test -race ./... -timeout=5m` - passed
- `GOTOOLCHAIN=go1.26.4 go vet ./...` - passed
- `golangci-lint v2.12.2 run --timeout=5m` - passed
- `DOCKER_BUILDKIT=1 GOTOOLCHAIN=go1.26.4 go test -tags=acceptance -p 1 -race acceptance_test.go -run '^TestCore/apk/amd64/complex$'` - passed
- `apk add --allow-untrusted repro-tree_1.0.0_x86_64.apk` on Alpine 3.20 - failed before with `v2 package format error`; passed after with the payload installed

### DIFF
--- a/apk/apk.go
+++ b/apk/apk.go
@@ -36,6 +36,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"io/fs"
 	"net/mail"
 	"os"
 	"strings"
@@ -419,7 +420,7 @@ func createFilesInsideTarGz(info *nfpm.Info, tw *tar.Writer, sizep *int64) (err 
 		case files.TypeDir, files.TypeImplicitDir:
 			err = tw.WriteHeader(&tar.Header{
 				Name:     file.Destination,
-				Mode:     int64(file.FileInfo.Mode),
+				Mode:     int64(normalizeFileMode(file.FileInfo.Mode)),
 				Typeflag: tar.TypeDir,
 				Uname:    file.FileInfo.Owner,
 				Gname:    file.FileInfo.Group,
@@ -453,9 +454,9 @@ func copyToTarAndDigest(file *files.Content, tw *tar.Writer, sizep *int64) error
 		return err
 	}
 
-	// tar.FileInfoHeader only uses file.Mode().Perm() which masks the mode with
-	// 0o777 which we don't want because we want to be able to set the suid bit.
-	header.Mode = int64(file.Mode())
+	// Preserve Unix special bits without leaking Go's file type flags into the
+	// tar mode field.
+	header.Mode = int64(normalizeFileMode(file.Mode()))
 	header.Name = files.AsRelativePath(file.Destination)
 	header.Uname = file.FileInfo.Owner
 	header.Gname = file.FileInfo.Group
@@ -465,6 +466,20 @@ func copyToTarAndDigest(file *files.Content, tw *tar.Writer, sizep *int64) error
 
 	*sizep += file.Size()
 	return nil
+}
+
+func normalizeFileMode(mode fs.FileMode) fs.FileMode {
+	result := mode & 0o7777
+	if mode&fs.ModeSetuid != 0 {
+		result |= 0o4000
+	}
+	if mode&fs.ModeSetgid != 0 {
+		result |= 0o2000
+	}
+	if mode&fs.ModeSticky != 0 {
+		result |= 0o1000
+	}
+	return result
 }
 
 // reference: https://wiki.adelielinux.org/wiki/APK_internals#.PKGINFO
--- a/apk/apk_test.go
+++ b/apk/apk_test.go
@@ -642,7 +642,7 @@ func TestTreeDirectoryMode(t *testing.T) {
 	sourceInfo, err := os.Stat(source)
 	require.NoError(t, err)
 	require.Equal(t, byte(tar.TypeDir), header.Typeflag)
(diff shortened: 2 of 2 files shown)


caarlos0 (approved):

caarlos0:
thanks for the PR!

windows will need some fixes I think, though

ychampion:
Fixed. The assertion now applies the package umask to the source permissions; the focused test and the full APK package under the race detector both pass.

caarlos0 (approved):

caarlos0:
Thank you!

PR #1118, opened 2026-07-22
wip: feat(rpm): Implement RPM triggers.

erikgeiser:
This PR adds support for RPM triggers, which is a concept similar to `deb` triggers which I implemented in #178. It should be complete, but I'm not completely sure about `srpm`s yet and I want to test this branch in practice before I remove the WIP.

### DIFF
--- a/rpm/coverage_test.go
+++ b/rpm/coverage_test.go
@@ -92,6 +92,43 @@ func TestApplyRelationsInvalidPostRequire(t *testing.T) {
 	require.Contains(t, err.Error(), "unknown version operator")
 }
 
+func TestReadTriggersErrors(t *testing.T) {
+	for name, trigger := range map[string]struct {
+		trigger  nfpm.RPMTrigger
+		expected string
+	}{
+		"type":    {nfpm.RPMTrigger{Type: "invalid", Script: "script", Conditions: []string{"foo"}}, "unknown trigger type"},
+		"script":  {nfpm.RPMTrigger{Type: "in", Conditions: []string{"foo"}}, "script must be provided"},
+		"missing": {nfpm.RPMTrigger{Type: "in", Script: "../testdata/does-not-exist.sh", Conditions: []string{"foo"}}, "does-not-exist.sh"},
+	} {
+		t.Run(name, func(t *testing.T) {
+			info := &nfpm.Info{Overridables: nfpm.Overridables{RPM: nfpm.RPM{Triggers: []nfpm.RPMTrigger{trigger.trigger}}}}
+
+			_, err := readTriggers(info)
+			require.ErrorContains(t, err, trigger.expected)
+		})
+	}
+
+	t.Run("condition", func(t *testing.T) {
+		info := &nfpm.Info{Overridables: nfpm.Overridables{RPM: nfpm.RPM{Triggers: []nfpm.RPMTrigger{{
+			Type: "in", Script: "../testdata/scripts/postinstall.sh", Conditions: []string{"foo >>> 2"},
+		}}}}}
+
+		_, err := readTriggers(info)
+		require.ErrorContains(t, err, "invalid condition")
+	})
+
+	t.Run("no conditions", func(t *testing.T) {
+		info := &nfpm.Info{Overridables: nfpm.Overridables{RPM: nfpm.RPM{Triggers: []nfpm.RPMTrigger{{
+			Type: "in", Script: "../testdata/scripts/postinstall.sh",
+		}}}}}
+
+		triggers, err := readTriggers(info)
+		require.NoError(t, err)
+		require.Empty(t, triggers[0].conditions)
+	})
+}
+
 func TestRenderSpecChangelog(t *testing.T) {
 	t.Run("renders newest first", func(t *testing.T) {
 		info := exampleInfo()
--- a/rpm/spec.go
+++ b/rpm/spec.go
@@ -98,6 +98,12 @@ func generateSpec(info *nfpm.Info, sourceName string) (string, error) {
 	writeScriptSection(&b, "posttrans", scripts.postTrans)
 	writeScriptSection(&b, "verifyscript", scripts.verify)
 
+	triggers, err := readTriggers(info)
+	if err != nil {
+		return "", err
+	}
+	writeTriggerSections(&b, triggers)
+
 	b.WriteString("\n%files\n")
 	writeFilesSection(&b, info)
 
@@ -118,6 +124,29 @@ func writeScriptSection(b *strings.Builder, section, body string) {
 	fmt.Fprintf(b, "\n%%%s\n%s", section, body)
 	if !strings.HasSuffix(body, "\n") {
 		b.WriteString("\n")
+	}
+}
+
+func writeTriggerSections(b *strings.Builder, triggers []trigger) {
+	for _, trigger := range triggers {
+		if len(trigger.conditions) == 0 {
+			continue
+		}
+
+		conditions := make([]string, 0, len(trigger.conditions))
+		for _, condition := range trigger.conditions {
+			conditions = append(conditions, escapeSpecText(condition.String()))
+		}
+
+		fmt.Fprintf(b, "\n%%trigger%s", trigger.directive)
+		if trigger.interpreter != "/bin/sh" {
+			fmt.Fprintf(b, " -p %s", escapeSpecText(trigger.interpreter))
+		}
+
+		fmt.Fprintf(b, " -- %s\n%s", strings.Join(conditions, ", "), escapeSpecText(trigger.script))
+		if !strings.HasSuffix(trigger.script, "\n") {
+			b.WriteString("\n")
+		}
 	}
 }
 
--- a/rpm/triggers.go
+++ b/rpm/triggers.go
@@ -1,0 +1,96 @@
+package rpm
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/goreleaser/nfpm/v2"
+	"go.digitalxero.dev/rpm"
+)
+
+type trigger struct {
+	triggerType rpm.TriggerKind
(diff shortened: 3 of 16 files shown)


djgilcrease (approved):

rpm/triggers.go
```diff
@@ -0,0 +1,96 @@
...
+
+		conditions := make([]rpm.Relation, 0, len(t.Conditions))
+		for _, condition := range t.Conditions {
+			relation, err := rpm.ParseRelation(condition)
+			if err != nil {
+				return nil, fmt.Errorf("rpm trigger %d: invalid condition %q: %w", i+1, condition, err)
+			}
+
+			conditions = append(conditions, relation)
+		}
```
caarlos0, line 44: An empty `conditions` list is accepted here and is then dropped further down. The library removes every trigger that has no condition or an empty body (`writeTriggers` in `header.go`: `if len(tr.conds) == 0 || tr.script == ""`), and `rpm/spec.go:132` removes the no-condition case again. nfpm shows success and makes a package that has no trigger.

**Trigger:** `rpm.triggers: [{type: in, script: ./x.sh}]` with no `conditions`, or with an `x.sh` that is empty.
**Impact:** the configuration goes away without a message. RPM has no trigger without a condition, thus this is always a user error.
**Smallest fix:** after this loop, return `fmt.Errorf("rpm trigger %d: at least one condition must be provided", i+1)` when `len(conditions) == 0`, and do the same for an empty `body`. This one check corrects the two output paths and lets the `continue` at `rpm/spec.go:132` go away.
**Test:** make the `no conditions` case at `rpm/coverage_test.go:121` an error case, and add an empty-script case that uses `t.TempDir()`.

Also correct `www/content/docs/configuration.md:369`, which makes this silent removal a documented rule.

rpm/spec.go
```diff
@@ -121,6 +127,29 @@ func writeScriptSection(b *strings.Builder, section, body string) {
 	}
 }
 
+func writeTriggerSections(b *strings.Builder, triggers []trigger) {
+	for _, trigger := range triggers {
+		if len(trigger.conditions) == 0 {
+			continue
+		}
```
caarlos0, line 134: This removes a trigger that has no conditions, but not one that has an empty script body. The library removes the two (`writeTriggers` in `header.go`: `len(tr.conds) == 0 || tr.script == ""`).

**Trigger:** a `script:` file that exists and has zero bytes.
**Impact:** the binary RPM has no trigger, but the generated spec writes a `%triggerin -- target` section that has an empty body. One configuration makes two different packages. This is against the rule in `rpm/srpm.go:27` that `rpmbuild --rebuild` gives the binary RPM again.
**Smallest fix:** refuse the two cases in `readTriggers` (see the comment at `rpm/triggers.go:44`) and remove this `continue` completely.
**Test:** a case in `rpm/srpm_test.go` that shows the spec has no `%trigger` section, and a case in `rpm/rpm_test.go` where `pkg.Header.GetStrings(tagTriggerName)` finds no tag.

rpm/spec.go
```diff
@@ -98,6 +98,12 @@ func generateSpec(info *nfpm.Info, sourceName string) (string, error) {
 	writeScriptSection(&b, "posttrans", scripts.postTrans)
 	writeScriptSection(&b, "verifyscript", scripts.verify)
 
+	triggers, err := readTriggers(info)
+	if err != nil {
+		return "", err
+	}
+	writeTriggerSections(&b, triggers)
```
caarlos0, line 105: This is the part of the change that has no true validation, and it agrees with your doubt about SRPMs in the description. All the text that `writeTriggerSections` writes (`%triggerprein`, `-p <interpreter>`, the condition list with commas, the `%%` escape sequences) is only compared as a string by `require.Contains` at `rpm/srpm_test.go:144`. No test gives this text to `rpmbuild`.

**Trigger:** any error in the spec syntax here makes a `.src.rpm` that the user cannot rebuild, and no test fails.
**Impact:** a defective SRPM goes to the user.
**Smallest fix:** the necessary equipment is already available. Add an `rpm.triggers` block to `testdata/acceptance/srpm.rebuild.yaml` that has the four types, one non-default interpreter, one default interpreter and one condition that has a version. The `rpmbuild --rebuild` at `testdata/acceptance/srpm.dockerfile:24` then reads it for real.
**Test:** after the rebuild, add `RUN rpm -qp --triggers /tmp/foo.rpm | grep -F 'triggerin'` and the equivalent for the other three types.

This also gives coverage to the `trigger.interpreter != "/bin/sh"` condition on line 142. Its no-`-p` branch is never used by a test now, because `TestSRPMSpecContents` only uses `/bin/bash`.

rpm/coverage_test.go
```diff
@@ -92,6 +92,43 @@ func TestApplyRelationsInvalidPostRequire(t *testing.T) {
...
+
+	t.Run("no conditions", func(t *testing.T) {
+		info := &nfpm.Info{Overridables: nfpm.Overridables{RPM: nfpm.RPM{Triggers: []nfpm.RPMTrigger{{
+			Type: "in", Script: "../testdata/scripts/postinstall.sh",
+		}}}}}
+
+		triggers, err := readTriggers(info)
+		require.NoError(t, err)
+		require.Empty(t, triggers[0].conditions)
+	})
```
caarlos0, line 129: This case examines an internal field and not the result that the user gets, and it makes the silent removal permanent. A trigger that has no conditions never goes into the package, thus the important assertion is that the package has no trigger, and not that `conditions` is empty.

Make this an error case together with the correction at `rpm/triggers.go:44`. The name of the test, `TestReadTriggersErrors`, then becomes correct again.

testdata/acceptance/rpm.dockerfile
```diff
@@ -234,3 +234,26 @@ FROM min AS postrequires
...
+      'RPM trigger target' \
+      '%files' \
+      > /tmp/rpmbuild/SPECS/trigger-target.spec && \
+    rpmbuild --define '_topdir /tmp/rpmbuild' -bb /tmp/rpmbuild/SPECS/trigger-target.spec
+RUN rpm -ivh /tmp/rpmbuild/RPMS/noarch/trigger-target-1.0-1.noarch.rpm
+RUN grep -qx prein /tmp/rpm-trigger-proof
+RUN grep -qx in /tmp/rpm-trigger-proof
+RUN rpm -e trigger-target
+RUN grep -qx un /tmp/rpm-trigger-proof
+RUN grep -qx postun /tmp/rpm-trigger-proof
```
caarlos0, line 259: The file `/tmp/rpm-trigger-proof` only gets more lines and is never made empty again, and there is no negative control. Thus the last two lines can be successful for the wrong reason: if a defect made the four triggers type `in`, the `rpm -ivh` on line 254 would write the four lines, and lines 258 and 259 would still be successful after the erase.

**Smallest fix:** two more assertions.
- `RUN test ! -e /tmp/rpm-trigger-proof` immediately after line 239. This shows that no trigger operates while the target package is absent.
- `RUN ! grep -qxE 'un|postun' /tmp/rpm-trigger-proof` immediately before line 257. This shows that the removal triggers did not operate too early.

The remainder of this stage is good work. To build a true target package and to drive the four types through a true rpm is correct.

www/content/docs/configuration.md
```diff
@@ -358,6 +358,19 @@ rpm:
...
 
+  # RPM package triggers.
+  triggers:
+    # The supported types are prein, in, un, and postun.
+    - type: in
+      script: ./scripts/trigger-in.sh
+      # Defaults to /bin/sh.
+      interpreter: /bin/bash
+      # Optional. Conditions use RPM dependency syntax and may contain multiple entries.
+      # No trigger is added when this list is empty.
```
caarlos0, line 369: `No trigger is added when this list is empty.` makes the silent removal of user configuration a documented rule. See the comment at `rpm/triggers.go:44`. When `conditions` becomes necessary, remove this line and also remove `omitempty` from the `Conditions` field in `nfpm.go`, so that the JSON schema shows the field as required.

While you are in this file: line 363 gives the four type names but not their function. One clause for each type helps the user: `prein` and `in` operate before and after the installation of a package that agrees with a condition, `un` and `postun` operate before and after its removal.

caarlos0 (approved):
## Review

The design is right and the binary RPM path is solid. `readTriggers` validates and normalizes, `applyTriggers` maps cleanly onto the builder, the config shape (`type`/`script`/`interpreter`/`conditions`) agrees with the existing `rpm.scripts` block, the new field is additive and `omitempty`, and `TestRPMTriggers` locks the tags, the flag bits and the per-condition index fan-out that nfpm actually owns. The `triggers` acceptance stage builds a real target package and drives all four types through a real rpm. That is the correct way to test this.

Three items to close before this leaves WIP. Each one has an inline comment.

1. An empty `conditions` list, and a `script` file that is empty, are accepted and then dropped without a message. Refuse them in `readTriggers`.
2. That same empty-script case makes the binary RPM and the generated spec disagree. This breaks the rule in `rpm/srpm.go:27` that `rpmbuild --rebuild` gives the binary RPM again.
3. The generated `%trigger*` spec text never goes through `rpmbuild`. This is the SRPM doubt from your description, and the correction is small: put triggers into `testdata/acceptance/srpm.rebuild.yaml`.

Out of scope, but recorded: a rich condition such as `(foo if bar)` loses the `RPMSENSE_RICH` bit, because `go.digitalxero.dev/rpm@v0.2.1` has no such constant. `depends:` already uses the same `ParseRelation` at `rpm/relations.go:30`, thus the condition is pre-existing and applies to the full repository. Do not correct it here.

Verification gap: I did not build the code, run the test suite or run the docker acceptance stages. All the findings come from reading the change, the dependency source and the adjacent code.

I approve, because no item stops the approach. Please complete items 1 to 3 before you merge.

Disclosure: an automated reviewer (a bot) made this review.

PR #1131, opened 2026-08-29
fix: Include optional dependencies in Arch packages

### LINKED ISSUE #1123: archlinux: the `optdepend` info was not generated
### What happened?

Hello, I tried adding the `recommends` or `suggests` options to my configuration, but the generated Arch Linux packages do not include `optdepend` information. I couldn't find any other similar options in the documentation either. I'm not sure if there's an error in my configuration or if this feature isn't supported at all.

Example Config:
```yaml
name: foo
arch: amd64
version: 1.2.3
maintainer: Carlos Alexandro Becker <root@carlosbecker.com>
provides:
  - bar
depends:
  - git
recommends:
  - golang
# suggests:
#   - bzr
```

---

Commands Executed:
```bash
nfpm package -f nfpm-test.yaml -p archlinux
```

---

Generated Package Info:
```
# Generated by nfpm
arch = x86_64
builddate = 1786113210
packager = Unknown Packager
pkgbase = foo
pkgdesc = no description given
pkgname = foo
pkgver = 1.2.3-1
size = 0
provides = bar
depend = git
```

### nfpm version

```bash
> nfpm --version
_____ ____  __  __
_ __ |  ___|  _ \|  \/  |
| '_ \| |_  | |_) | |\/| |
| | | |  _| |  __/| |  | |
|_| |_|_|   |_|   |_|  |_|
nfpm: a simple and 0-dependencies apk, arch linux, deb, ipk, msix, and rpm packager written in Go
https://nfpm.goreleaser.com

GitVersion:    2.47.0
GitCommit:     40c7c8f7376400e9464d04a2a099045cf5598f8e
GitTreeState:  false
BuildDate:     2026-06-20T23:17:25Z
BuiltBy:       goreleaser
GoVersion:     go1.26.4
Compiler:      gc
ModuleSum:     h1:0bioJAjWaMPntgDqynP4ze0Wt4zYqYSFJ5/BBy9XIGI=
Platform:      linux/amd64
```

### Search

- [x] I did search for other open and closed issues before opening this.

### Code of Conduct

### Additional context

_No response_

fredrikblau:
### What changed

Arch packages now include `recommends` and `suggests` entries as `optdepend` metadata in `.PKGINFO`.

### Verification

- `go test ./...`
- `go build ./cmd/nfpm`
- `go vet ./...`
- `git diff --check`

Closes #1123


arch/arch.go
```diff
@@ -359,6 +359,13 @@ func createPkginfo(info *nfpm.Info, tw *tar.Writer, totalSize int64) (*MtreeEntr
 		}
 	}
 
+	for _, optdepend := range append(info.Recommends, info.Suggests...) {
```
caarlos0, line 362: Nit, non-blocking: `append(info.Recommends, ...)` writes into the caller's backing array when `cap > len`. This can occur, because `expandEnvVarsStringSlice` uses `slices.Delete`, which keeps the capacity after it removes an empty entry (`recommends: [git, "${UNSET}"]` gives len 1, cap 2).

I could not build a repro that a user can observe: all writes land past `len`, and no code in this repository reads past `len` of `Recommends`. So this is style, not a bug. `slices.Concat` removes the question:

```go
for _, optdepend := range slices.Concat(info.Recommends, info.Suggests) {
```

No new test is necessary.

arch/arch_test.go
```diff
@@ -161,6 +167,8 @@ func TestArchPkginfo(t *testing.T) {
 	require.Equal(t, "zsh", fields["conflict"])
 	require.Equal(t, "bzr", fields["provides"])
 	require.Equal(t, "bash", fields["depend"])
+	require.Contains(t, string(pkginfoData), "optdepend = git\n")
+	require.Contains(t, string(pkginfoData), "optdepend = fish\n")
```
caarlos0, line 171: These two assertions read the raw bytes, because `extractPkginfoFields` (line 248) puts each key in a `map[string]string` and thus keeps only the last value of a repeated key. Result: no test can find a dropped, duplicated, or wrongly repeated `optdepend`, and this is the first field in `.PKGINFO` that occurs more than one time in the example data.

A small helper keeps the assertion at the same layer as the others and makes the count part of the contract:

```go
func extractPkginfoValues(data []byte, key string) []string {
	var out []string
	for _, line := range strings.Split(strings.TrimSpace(string(data)), "\n") {
		if k, v, ok := strings.Cut(line, " = "); ok && k == key {
			out = append(out, v)
		}
	}
	return out
}
```

Then:

```go
require.ElementsMatch(t, []string{"git", "fish"}, extractPkginfoValues(pkginfoData, "optdepend"))
```

`ElementsMatch` prints both slices on failure and does not make the line order a contract.

caarlos0 (approved):
Approving. The mapping of `recommends` + `suggests` to `optdepend` is correct for `.PKGINFO`, it follows the existing loop pattern in `createPkginfo`, and it adds no configuration surface. The two comments below are non-blocking.

**Acceptance coverage (no single site in this diff)**

`testdata/acceptance/core.complex.yaml` already sets `recommends: fish` and `suggests: zsh`, and the `complex` stage of `testdata/acceptance/archlinux.dockerfile` already greps `pacman -Qi` for `Depends On`, `Replaces` and `Provides`. Nothing there asserts the new field, so the unit test is the only proof. One more line in that stage makes pacman itself prove that it parses the new entry:

```dockerfile
RUN pacman -Qi foo | grep "Optional Deps\\s*: fish"
```

I verified that no acceptance stage does strict output matching, so this change breaks no current stage.

**Verification gaps**

- I did not build or run tests locally, as requested. `exampleInfo()` is shared by 12 tests in `arch/arch_test.go`, so `go test ./arch/...` is the check that matters.
- I did not run the independent `caarlos0` and `anvil` sub-agent checks. The diff is 7 lines and fits one direct read.

_This review is from an automated (bot) reviewer._

fredrikblau:
Thanks for the review.

Added the acceptance assertion you suggested to the `complex` stage of `testdata/acceptance/archlinux.dockerfile`, right after the existing `Provides` check:

```dockerfile
RUN pacman -Qi foo | grep "Optional Deps\\s*: fish"
```

`core.complex.yaml` sets `recommends: fish` and `suggests: zsh`, and `optdepend` entries are written recommends-first, so `fish` is the first line pacman prints under `Optional Deps`. I have no Docker available locally, so that stage is verified by CI rather than by me.

Also rebased on current `main`. The previous run's `ruleguard / scan` failure was `internal error: package "maps" without types was imported from ...internal/maps`, which is unrelated to this diff and no longer reproduces on `main`; the rebase should clear it.

`go test ./arch/...` passes on the rebased branch.

